PR #233, opened 2023-12-14
Update README.md

arctiqmk2:
Updated the oxide provider definition in the example.  The greater-than-or-equal on the oxide provider definition causes the terraform init to fail with a "Error: Failed to query available provider packages" at this time.  Updating the example.  This may change as new versions are released.


karencfv (approved):
Hi there @arctiqmk2, thanks a bunch for taking the time to fix the docs :bowing_woman:

PR #269, opened 2024-01-30
Select SSH keys to be added to instance on create

karencfv:
Depends on https://github.com/oxidecomputer/omicron/pull/4764


karencfv:
Just need to test this on dogfood and it's good to go

PR #285, opened 2024-03-04
Run acceptance tests as part of CI

karencfv:


karencfv:
For now acceptance tests will only run on releases due to secrets not working on forks.

PR #304, opened 2024-04-09
Address Lot Datasource and Resource

### LINKED ISSUE #306: Address Lots

internet-diglett:
Summary
---
Add the ability to manage address lots via Terraform Provider

Changes
---
- [x] Add datasource for Address Lots
- [ ] Add resource for Address Lots
- [ ] Update changelog
- [ ] Update documentation

Related
---
Closes #306

### DIFF
--- a/internal/provider/data_source_address_lots_test.go
+++ b/internal/provider/data_source_address_lots_test.go
@@ -23,7 +23,7 @@ data "oxide_networking_address_lots" "{{.BlockName}}" {
 }
 `
 
-func TestAccCloudDataSourceNetworkingAddressLots_full(t *testing.T) {
+func TestAccSiloDataSourceNetworkingAddressLots_full(t *testing.T) {
 	blockName := newBlockName("datasource-networking-address-lots")
 	config, err := parsedAccConfig(
 		dataSourceNetworkingAddressLotConfig{


internal/provider/data_source_address_lot.go
karencfv, line 1: Let's change the name of the file to data_source_address_lots.go so it's consistent with the data source name :)

internal/provider/data_source_address_lot_test.go
karencfv, line 1: Same about the name change on this file

internal/provider/data_source_address_lot.go
```diff
@@ -0,0 +1,143 @@
...
+				Computed: true,
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"id": schema.StringAttribute{
+							Computed:    true,
+							Description: "Address Lot ID",
+						},
+						"description": schema.StringAttribute{
+							Computed:    true,
+							Description: "human-readable free-form text about an Address Lot",
```
karencfv, line 77:
```suggestion
							Description: "Human-readable free-form text about an Address Lot",
```

internal/provider/data_source_address_lot_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+var datasourceNetworkingAddressLotsConfigTpl = `
+data "oxide_networking_address_lots" "{{.BlockName}}" {
+	timeouts = {
+	  read = "1m"
+	}
+}
+`
+
+func TestAccCloudDataSourceNetworkingAddressLots_full(t *testing.T) {
```
karencfv, line 26: The prefix is to differentiate between "cloud" and "silo" endpoints. I use this manly for testing on the colo rack, where the user on the GH action doesn't have admin privileges

```suggestion
func TestAccSiloDataSourceNetworkingAddressLots_full(t *testing.T) {
```

karencfv (commented):
Thanks for opening up this PR! Super excited to have more people contributing to this repo.

Looks good overall, just a few comments:

- Could we have a PR per datasource/resource pair? We could have a meta issue to track each of the checklist items you have on the description of this PR. This way it'll be easier for me to review, and will make the commit history and changelog more comprehensible :)
- Talking about the changelog, could you add an entry to the `.changelog/0.4.0` file? I have a little generator that reads the file to create these nice changelog [entries](https://github.com/oxidecomputer/terraform-provider-oxide/blob/main/CHANGELOG.md#020-2024feb13). Here's a good example from the previous release https://github.com/oxidecomputer/terraform-provider-oxide/blob/main/.changelog/0.2.0.toml
- Could we add some documentation? I like to add the documentation at the same time we develop the resources/datasources so it doesn't become cumbersome to add docs later down the track :sweat_smile: Here's an example https://github.com/oxidecomputer/terraform-provider-oxide/blob/main/docs/data-sources/oxide_projects.md

internet-diglett:
Looks like we may need to update existing endpoints / expose some new endpoints from the control plane to implement Address Lots in a similar way to how we handle similar resources, like IP Pools:

https://docs.oxide.computer/api/networking_address_lot_list

https://docs.oxide.computer/api/ip_pool_list

Essentially with address lots, we cannot create a lot, then create a block for a lot. We have to declare the lot and blocks at the same time. However, you cannot _view_ the blocks when you view a lot. You cannot create / view / remove lots either. Conversely, you can create pools, then create / view / remove ranges for a pool separately.

Will discuss with the networking team and revisit once a decision has been made.

internet-diglett:
Confirmed, we're going to update the system networking APIs to make them more consistent, then resume updating the Terraform provider.

PR #331, opened 2024-08-12
Update README to refer to credentials.toml

wfchandler:
We no longer use `hosts.toml`. Update README to use the new file.


README.md
```diff
@@ -15,7 +15,7 @@ To generate a token, follow these steps:
 - Make sure you have installed the Oxide CLI
 - Log in via the Oxide console.
 - Run `oxide auth login --host <host>`
-- Run `oxide auth status --host <host> --show-token` to retrieve the token (host and credential information is saved in `$HOME/.config/oxide/hosts.toml`).
+- Run `oxide auth status --host <host> --show-token` to retrieve the token (host and credential information is saved in `$HOME/.config/oxide/credentials.toml`).
```
karencfv, line 18: Hmmm, I think the `--host` and `--show token` flags were removed from the `oxide auth status` command. Perhaps this sentence can be changed to tell the users that the tokens can be found in `$HOME/.config/oxide/credentials.toml` instead?

karencfv (commented):
Nice catch! Thanks @wfchandler 🙇‍♀️

README.md
ahl: Agreed; @karencfv what about if I implement https://github.com/oxidecomputer/oxide.rs/issues/596 and we change the instructions to be `oxide --profile foo env terraform plan` or whatever -- that's kind of what I had in mind for that, but I wanted to validate that we think that's the right approach before going that way.

karencfv: Thanks @ahl! 

On the terraform side I was thinking about adding a `profile` argument to the provider block (I should open up an issue 😄 ) like this:

```
provider "oxide" {
  profile = "prod"
}
```

But it would also be really useful to have the command you mention too! I can see it being very practical for people wanting to apply the same terraform plan on several silos without having to go and manually edit the tf files.

So, yes. If you have the time, it'd be really nice to have that command in addition to what I'm planning on the TF side!

ahl: I might have my bit set the wrong way here, but I don't think this should be the domain of the terraform provider? That might be wrong or inconsistent (in that the Rust SDK is profile-aware), but my inclination is that I wouldn't ever put a profile name into a tf config: I'm probably going to check in the tf config to a git repo and I don't think I'd want everyone to be **required** to have the same profile name. What do you think?

karencfv: Let's move this conversation over to https://github.com/oxidecomputer/terraform-provider-oxide/issues/334 ! I'll respond there to have a better record of this discussion :)

wfchandler:
Thanks @karencfv, I've updated to remove the reference to `--show-token`.

karencfv (approved):
Thanks for the fix!

PR #367, opened 2024-11-18
codeowners: update owners

sudomateo:
This patch adds Matthew Sanabria (`@sudomateo`) to `CODEOWNERS` and removes those who have served their time.


karencfv (approved):
@augustuswm @sudomateo Thanks for dealing with all the provider emails these past years 😄

Issue #378, opened 2025-01-08
Support in-place instance ephemeral and floating IP changes


askfongjojo:
### Overview

Currently, a change in the instance external IP spec results in a forced replacement in `terraform apply`, e.g.
```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # oxide_instance.mysql[0] must be replaced
-/+ resource "oxide_instance" "mysql" {
      ~ external_ips       = [
          - { # forces replacement
              - id   = "6c75764b-0350-4dfe-9f13-920ae0c3e75f" -> null
              - type = "floating" -> null
            },
            # (1 unchanged element hidden)
        ]
      ~ id                 = "8bc31b24-22c7-4d40-82aa-22a603564abe" -> (known after apply)
        name               = "mysql-0"
      ~ network_interfaces = [
          - {
              - description   = "Default NIC" -> null
              - id            = "93193c53-74f9-4de3-8311-0a788a1706c6" -> null
              - ip_address    = "172.30.0.11" -> null
              - mac_address   = "A8:40:25:F0:00:06" -> null
              - name          = "default" -> null
              - primary       = true -> null
              - subnet_id     = "11a7986f-475d-4bf1-bb96-bacb3dbeec3a" -> null
              - time_created  = "2025-01-08 19:50:49.67032 +0000 UTC" -> null
              - time_modified = "2025-01-08 19:50:49.67032 +0000 UTC" -> null
              - vpc_id        = "cd36ca5d-b536-4316-8b26-e330cafa3b9f" -> null
            },
          + {
              + description   = "Default NIC"
              + id            = (known after apply)
              + ip_address    = (known after apply)
              + mac_address   = (known after apply)
              + name          = "default"
              + primary       = (known after apply)
              + subnet_id     = "11a7986f-475d-4bf1-bb96-bacb3dbeec3a"
              + time_created  = (known after apply)
              + time_modified = (known after apply)
              + vpc_id        = "cd36ca5d-b536-4316-8b26-e330cafa3b9f"
            },
        ]
      ~ time_created       = "2025-01-08 19:50:49.429581 +0000 UTC" -> (known after apply)
      ~ time_modified      = "2025-01-08 19:50:49.429581 +0000 UTC" -> (known after apply)
        # (8 unchanged attributes hidden)
    }

Plan: 1 to add, 0 to change, 1 to destroy.
[...]

PR #381, opened 2025-01-09
oxide_instance: allow in-place `external_ips` modification

sudomateo:
Closes https://github.com/oxidecomputer/terraform-provider-oxide/issues/378.

This patch updates the `oxide_instance` resource to support in-place modifications to its `external_ips` attribute. That is, the instance will no longer be destroyed and recreated when `external_ips` is modified. Instead, the instance will be stopped, external IPs will be detached and/or attached, and the instance will be started.

This patch also updates the `Read` method to refresh the external IPs attached to the instance. Previously, since the external IPs required a resource replacement there was no need to refresh them.

There are 2 types of external IPs, ephemeral and floating. There can be at most 1 ephemeral external IP attached to an instance and any number of floating external IPs. That means in order to modify ephemeral external IPs the external IPs must be detached first and then attached.

This patch needs the following work before it can be merged.
- [x] Add tests for the new `external_ips` modification logic.
- [x] Add validation logic to `external_ips` to enforce at most 1 ephemeral external IP. This was previously "validated" during instance creation but is not updated during instance update.

Here's the error that's returned when an instance is created with more than 1 ephemeral external IP.

```
oxide_instance.foo: Creating...
╷
│ Error: Error creating instance
│
│   with oxide_instance.foo,
│   on main.tf line 24, in resource "oxide_instance" "foo":
│   24: resource "oxide_instance" "foo" {
│
│ API error: POST https://oxide.example.com/v1/instances?project=5476ccc9-464d-4dc4-bfc0-5154de1c986f
│ ----------- RESPONSE -----------
│ Status: 400 InvalidRequest
│ Message: An instance may not have more than 1 ephemeral IP address
│ RequestID: fc6144e5-fa76-4583-a024-2e49ce17140e
│ ------- RESPONSE HEADERS -------
│ Content-Type: [application/json]
│ X-Request-Id: [fc6144e5-fa76-4583-a024-2e49ce17140e]
│ Date: [Thu, 09 Jan 2025 03:28:48 GMT]
│ Content-Length: [166]
│
╵
```

### DIFF
--- a/internal/provider/resource_instance.go
+++ b/internal/provider/resource_instance.go
@@ -21,6 +21,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/boolplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/setplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringdefault"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
@@ -34,45 +35,6 @@ var (
 	_ resource.Resource              = (*instanceResource)(nil)
 	_ resource.ResourceWithConfigure = (*instanceResource)(nil)
 )
-
-var _ validator.List = externalIPValidator{}
-
-type externalIPValidator struct{}
-
-func (f externalIPValidator) Description(context.Context) string {
-	return "foo"
-}
-
-func (f externalIPValidator) MarkdownDescription(context.Context) string {
-	return "foo"
-}
-
-func (f externalIPValidator) ValidateList(ctx context.Context, req validator.ListRequest, resp *validator.ListResponse) {
-	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
-		return
-	}
-
-	var externalIPs []instanceResourceExternalIPModel
-	diags := req.ConfigValue.ElementsAs(ctx, &externalIPs, false)
-	resp.Diagnostics.Append(diags...)
-	if resp.Diagnostics.HasError() {
-		return
-	}
-
-	var ephemeralExternalIPs int
-	for _, externalIP := range externalIPs {
-		if externalIP.Type.ValueString() == string(oxide.ExternalIpCreateTypeEphemeral) {
-			ephemeralExternalIPs++
-		}
-	}
-	if ephemeralExternalIPs > 1 {
-		resp.Diagnostics.AddError(
-			fmt.Sprintf("Too many blocks with type = %q", oxide.ExternalIpCreateTypeEphemeral),
-			fmt.Sprintf("Only one block with type = %q is allowed, but found %d.", oxide.ExternalIpCreateTypeEphemeral, ephemeralExternalIPs),
-		)
-		return
-	}
-}
 
 // NewInstanceResource is a helper function to simplify the provider implementation.
 func NewInstanceResource() resource.Resource {
@@ -292,17 +254,19 @@ func (r *instanceResource) Schema(ctx context.Context, _ resource.SchemaRequest,
 					},
 				},
 			},
-			"external_ips": schema.ListNestedAttribute{
+			"external_ips": schema.SetNestedAttribute{
 				Optional:    true,
 				Description: "External IP addresses provided to this instance.",
-				Validators: []validator.List{
-					externalIPValidator{},
+				Validators: []validator.Set{
+					instanceExternalIPValidator{},
 				},
 				NestedObject: schema.NestedAttributeObject{
 					Attributes: map[string]schema.Attribute{
 						"id": schema.StringAttribute{
 							Description: "If type is ephemeral, ID of the IP pool to retrieve addresses from, or all available pools if not specified. If type is floating, ID of the floating IP",
 							Optional:    true,
+							Computed:    true,
+							Default:     stringdefault.StaticString(""),
 						},
 						"type": schema.StringAttribute{
 							Description: "Type of external IP.",
@@ -553,6 +517,9 @@ func (r *instanceResource) Read(ctx context.Context, req resource.ReadRequest, r
 			Type: types.StringValue(string(externalIP.Kind)),
 		})
 	}
+	if len(externalIPs) > 0 {
+		state.ExternalIPs = externalIPs
+	}
 
 	tflog.Trace(ctx, fmt.Sprintf("read instance with ID: %v", instance.Id), map[string]any{"success": true})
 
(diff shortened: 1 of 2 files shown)


internal/provider/resource_instance.go
```diff
@@ -256,11 +257,16 @@ func (r *instanceResource) Schema(ctx context.Context, _ resource.SchemaRequest,
...
+				Validators: []validator.Set{
+					instanceExternalIPValidator{},
+				},
 				NestedObject: schema.NestedAttributeObject{
 					Attributes: map[string]schema.Attribute{
 						"id": schema.StringAttribute{
 							Description: "If type is ephemeral, ID of the IP pool to retrieve addresses from, or all available pools if not specified. If type is floating, ID of the floating IP",
 							Optional:    true,
+							Computed:    true,
+							Default:     stringdefault.StaticString(""),
```
sudomateo, line 269: The `Computed` and `Default` accounts for the case where an instance created with an external IP using the default IP pool (i.e., `id = null`) would drift when read (e.g., `id = ""`) and require updating in place.

sudomateo:
I have one open question about this. This `oxide_instance` resource is currently coded to stop for any Update actions.

https://github.com/oxidecomputer/terraform-provider-oxide/blob/ac7cc52cc678a39afb860238504362ee91b55519/internal/provider/resource_instance.go#L586-L599

Since external IPs can be attached and detached from a running instance should we update the logic here to only stop the instance if there's an action that requires it?

Perhaps that would overly complicate this `Update` path though and it might be better to introduce some separate `oxide_external_ip_attachment` resource instead.

What do you think, @karencfv?

internal/provider/resource_instance.go
karencfv: Thanks! Do you mind adding that as a comment for future reference?

internal/provider/resource_instance.go
```diff
@@ -1020,6 +1061,33 @@ func newAttachedNetworkInterfacesModel(ctx context.Context, client *oxide.Client
 	return nicSet, nil
 }
 
+func newAttachedExternalIPModel(ctx context.Context, client *oxide.Client, instanceID string) (
+	[]instanceResourceExternalIPModel, diag.Diagnostics) {
+	var diags diag.Diagnostics
+	externalIPs := make([]instanceResourceExternalIPModel, 0)
+
+	// TODO: Support pagination when Go SDK supports it.
```
karencfv, line 1069: The Go SDK supports pagination in all endpoints that have it. In this case, it appears the endpoint doesn't have pagination https://github.com/oxidecomputer/omicron/blob/62614ab4abe2a4bc9976dd0e2fb395bc544413c3/openapi/nexus.json#L3623-L3668

I don't really know if it's in any one's roadmap to support pagination for this endpoint. Let's change the comment (or remove it, there's not much to do on our side tbh 🤷‍♀️ )

internal/provider/resource_instance.go
```diff
@@ -1183,6 +1251,115 @@ func deleteNICs(ctx context.Context, client *oxide.Client, models []instanceReso
...
+					fmt.Sprintf("Error attaching floating external IP with ID %s", externalIPID.ValueString()),
+					"API error: "+err.Error(),
+				)
+
+				return diags
+			}
+		default:
+			diags.AddError(
+				"Error attaching external IP",
+				fmt.Sprintf("Unknown external IP type %s", externalIPType.ValueString()),
```
karencfv, line 1301: I know it's unlikely anyone will hit this error, but the message is a little cryptic. Can we have something that's a little more descriptive?

internal/provider/resource_instance.go
```diff
@@ -1183,6 +1251,115 @@ func deleteNICs(ctx context.Context, client *oxide.Client, models []instanceReso
...
+					fmt.Sprintf("Error detaching floating external IP with ID %s", externalIPID.ValueString()),
+					"API error: "+err.Error(),
+				)
+
+				return diags
+			}
+		default:
+			diags.AddError(
+				"Error detaching external IP",
+				fmt.Sprintf("Unknown external IP type %s", externalIPType.ValueString()),
```
karencfv, line 1352: Same as above

internal/provider/resource_instance.go
```diff
@@ -1327,3 +1504,49 @@ func attrValueSliceContains(s []attr.Value, str string) (bool, error) {
...
+}
+
+// MarkdownDescription describes the validation in Markdown formatting.
+func (f instanceExternalIPValidator) MarkdownDescription(context.Context) string {
+	return "cannot have more than one ephemeral external ip"
+}
+
+// ValidateSet validates whether a set of [instanceExternalIPValidator] objects
+// has at most one object with type = "ephemeral".
+func (f instanceExternalIPValidator) ValidateSet(ctx context.Context, req validator.SetRequest, resp *validator.SetResponse) {
```
karencfv, line 1527: Oooohhh very nice!

karencfv (commented):
Thanks for working on this! Just a few nits, but otherwise looks great!

Since we can't run the acceptance tests on CI, do you mind pasting the results from running them locally?

Re: the stop-start question: Adding logic to stop/start only when necessary would certainly be the most correct way to go on about this change. That said, all updates without having to stop-start are in the roadmap for the hypervisor team. So the stopping and starting will go away eventually anyway. So I don't see this as a major issue to leave it as is for now. So, I'll leave it up top you! Time is a scarce resource right now, and I don't have strong feelings about this in particular.

sudomateo:
> Since we can't run the acceptance tests on CI, do you mind pasting the results from running them locally?

Done!

```
> TEST_ACC_NAME=TestAccCloudResourceInstance_extIPs make testacc
=== RUN   TestAccCloudResourceInstance_extIPs
=== PAUSE TestAccCloudResourceInstance_extIPs
=== CONT  TestAccCloudResourceInstance_extIPs
--- PASS: TestAccCloudResourceInstance_extIPs (42.95s)
PASS
ok      github.com/oxidecomputer/terraform-provider-oxide/internal/provider     42.957s
```

> Re: the stop-start question: Adding logic to stop/start only when necessary would certainly be the most correct way to go on about this change. That said, all updates without having to stop-start are in the roadmap for the hypervisor team. So the stopping and starting will go away eventually anyway. So I don't see this as a major issue to leave it as is for now. So, I'll leave it up top you! Time is a scarce resource right now, and I don't have strong feelings about this in particular.

Oh, I didn't know this! I'm also a bit pressed for time lately so I'm okay with leaving the behavior as-is and accepting the instance reboot for external IP attachment/detachment unless this happens to block a customer use case. If that comes up I think we can revisit this.

internal/provider/resource_instance.go
sudomateo: Updated! I was going to remove this block entirely but I just updated the error text. Let me know if it reads better.

internal/provider/resource_instance.go
sudomateo: I removed the comment. The context you provided was great, thank you!

internal/provider/resource_instance.go
sudomateo: Done!

karencfv (approved):
Looks great, thanks!

PR #383, opened 2025-01-14
adding profile lookup for credentials.toml

gurumachupalli:


internal/provider/provider.go
```diff
@@ -80,39 +86,55 @@ func (p *oxideProvider) Configure(ctx context.Context, req provider.ConfigureReq
...
+	if data.Profile.ValueString() != "" {
+		profile = data.Profile.ValueString()
 	}
 
-	if resp.Diagnostics.HasError() {
-		return
+	var config oxide.Config
+
+
+	if profile != "" {
```
karencfv, line 96: The SDK already has field validation for `Config`. This means we can directly pass the following without adding a conditional statement

```go
config := oxide.Config{
	Profile:   profile,
	Token:     token,
	Host:      host,
	UserAgent: fmt.Sprintf("terraform-provider-oxide/%s", Version),
}
```

karencfv (changes requested):
Hey @gurumachupalli, thanks a bunch for the contribution!

It'd be really nice to have some documentation to go along with this feature in our provider [docs](https://github.com/oxidecomputer/terraform-provider-oxide/blob/main/docs/index.md). Also, can we add an entry for our changelog [here](https://github.com/oxidecomputer/terraform-provider-oxide/blob/main/.changelog/0.7.0.toml)? You can see examples in our past [changelog](https://github.com/oxidecomputer/terraform-provider-oxide/blob/main/.changelog/) config files to see the format we follow.

Just a few minor bits, but overall looks great.

internal/provider/provider.go
gurumachupalli: removed the if and else logic since the validator is handling it and passing in profile to the sdk.

internal/provider/provider.go
```diff
@@ -56,6 +57,10 @@ func (p *oxideProvider) Schema(_ context.Context, _ provider.SchemaRequest, resp
 				Sensitive:   true,
 				Description: "Token used to authenticate",
 			},
+			"profile": schema.StringAttribute{
+				Optional:    true,
+				Description: "Profile used to authenticate. Retrieves host and token from credentials.toml",
```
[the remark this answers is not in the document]
gurumachupalli: Used the string validator as you mentioned, thanks!

internal/provider/provider.go
```diff
@@ -80,8 +105,11 @@ func (p *oxideProvider) Configure(ctx context.Context, req provider.ConfigureReq
...
 	}
+	if data.Profile.ValueString() != "" {
+		profile = data.Profile.ValueString()
+	}
 
-	if token == "" {
+	if token == "" && profile == "" {
 		resp.Diagnostics.AddError(
 			"Missing API Token Configuration",
 			"While configuring the provider, the API token was not found in "+
```
karencfv, line 115: Let's tweak this error message slightly

```go
"While configuring the provider, the API token was not found in "+
"the OXIDE_TOKEN environment variable, "+
"configuration block token attribute, or profile.",
```

internal/provider/provider.go
```diff
@@ -90,7 +118,7 @@ func (p *oxideProvider) Configure(ctx context.Context, req provider.ConfigureReq
 		)
 	}
 
-	if host == "" {
+	if host == "" && profile == "" {
 		resp.Diagnostics.AddError(
 			"Missing Host Configuration",
 			"While configuring the provider, the host was not found in "+
```
karencfv, line 124: Same as above
```go
"While configuring the provider, the host was not found in "+
"the OXIDE_HOST environment variable, "+
"configuration block host attribute, or profile.",
```

karencfv (changes requested):
Thanks for addressing my comments so quickly! Just a tiny change to the error messages and looks pretty much ready to go.

internal/provider/provider.go
gurumachupalli: done

internal/provider/provider.go
gurumachupalli: done

karencfv (commented):
Thanks @gurumachupalli! I've also tested manually and everything works as expected. Looks like the linter is unhappy 😄 Would you mind just running `gofmt -s -w internal/provider/provider.go` and committing the results?

gurumachupalli:
> Thanks @gurumachupalli! I've also tested manually and everything works as expected. Looks like the linter is unhappy 😄 Would you mind just running `gofmt -s -w internal/provider/provider.go` and committing the results?

pushed

karencfv (approved):
Thanks for the contribution!

gurumachupalli:
> Thanks for the contribution!

anytime! will have more in the future 😄 thanks for the feedback

Issue #392, opened 2025-02-12
Update examples to cover common use cases


sudomateo:
Every resource and data source contains example configuration that shows users how to use the respective resource. Many of these examples require additional configuration before they can be applied. This causes friction with users attempting to use Terraform/OpenTofu with Oxide. The examples for all resources and data sources should be updated to show complete configuration for common use cases.

For example, the `oxide_instance` resource has the following [example configuration](https://registry.terraform.io/providers/oxidecomputer/oxide/latest/docs/resources/oxide_instance#example-usage).

```hcl
resource "oxide_instance" "example" {
  project_id       = "c1dee930-a8e4-11ed-afa1-0242ac120002"
  boot_disk_id     = "611bb17d-6883-45be-b3aa-8a186fdeafe8"
  description      = "a test instance"
  name             = "myinstance"
  host_name        = "<host value>"
  memory           = 1073741824
  ncpus            = 1
  ssh_public_keys  = ["066cab1b-c550-4aea-8a80-8422fd3bfc40", "1aa748cb-26f0-4bf5-8faf-b202dc74d698"]
  disk_attachments = ["611bb17d-6883-45be-b3aa-8a186fdeafe8", "eb65d5cb-d8c5-4eae-bcf3-a0e89a633042"]
  network_interfaces = [
    {
      subnet_id   = "066cab1b-c550-4aea-8a80-8422fd3bfc40"
      vpc_id      = "9b9f9be1-96bf-44ad-864a-0dedae3b3999"
      description = "a sample nic"
      name        = "mynic"
    },
  ]
  timeouts = {
    read   = "1m"
    create = "3m"
    delete = "2m"
  }
}
```

This configuration has the following issues.

* The `boot_disk_id` refers to the ID of a _disk_, not an image. This is not intuitive for users coming from other cloud providers where they create instances from image IDs instead of from disks, even if the image ID is a disk under the hood. The example should be updated to use the `oxide_disk` resource as well.
* There are optional entries that detract from the example (e.g., `network_interfaces`, `timeouts`). These should be removed entirely or moved to other, more specific, examples so the primary example stays minimal and concise.
* It uses a mix of non-descriptive example values (e.g., `myinstance`) and placeholder values (i.e., `<host value>`). These should be updated with example values that one might commonly use in the real world (e.g., `app01`).

Applying these rules a better example would be the following.

```hcl
resource "oxide_disk" "app01" {
[...]

PR #393, opened 2025-02-13
docs: update oxide_instance resource

sudomateo:
This patch updates the example configuration and the attribute documentation for the `oxide_instance` resource as described in https://github.com/oxidecomputer/terraform-provider-oxide/issues/392.


docs/resources/oxide_instance.md
```diff
@@ -64,41 +71,27 @@ resource "oxide_instance" "example" {
...
-- `network_interfaces` (Set of Object, Optional) Virtual network interface devices attached to an instance. (see [below for nested schema](#nestedatt--nics))
-- `ssh_public_keys` (Set of String, Optional) An allowlist of IDs of the saved SSH public keys to be transferred to the instance via cloud-init during instance creation.
-- `start_on_create` (Boolean, Default `true`) Starts the instance on creation when set to true.
-- `timeouts` (Attribute, Optional) (see [below for nested schema](#nestedatt--timeouts))
-- `user_data` (String) User data for instance initialization systems (such as cloud-init). Must be a Base64-encoded string, as specified in [RFC 4648 § 4](https://datatracker.ietf.org/doc/html/rfc4648#section-4) (+ and / characters with padding). Maximum 32 KiB unencoded data.
+- `boot_disk_id` (String, Optional) ID of the disk to boot the instance from. When provided, this ID must also be present in `disk_attachments`.
+- `disk_attachments` (Set of String, Optional) IDs of the disks to be attached to the instance. The first entry in this set will be used to boot the instance unless `boot_disk_id` is provided.
```
karencfv, line 88: The set is an unordered list, so we can't really guarantee that the first entry will be used. The reason we're using a set instead of a list is because the instance disk list endpoint does not return the attached disks in the same order we set them.

docs/resources/oxide_instance.md
```diff
@@ -64,41 +71,27 @@ resource "oxide_instance" "example" {
...
-- `ssh_public_keys` (Set of String, Optional) An allowlist of IDs of the saved SSH public keys to be transferred to the instance via cloud-init during instance creation.
-- `start_on_create` (Boolean, Default `true`) Starts the instance on creation when set to true.
-- `timeouts` (Attribute, Optional) (see [below for nested schema](#nestedatt--timeouts))
-- `user_data` (String) User data for instance initialization systems (such as cloud-init). Must be a Base64-encoded string, as specified in [RFC 4648 § 4](https://datatracker.ietf.org/doc/html/rfc4648#section-4) (+ and / characters with padding). Maximum 32 KiB unencoded data.
+- `boot_disk_id` (String, Optional) ID of the disk to boot the instance from. When provided, this ID must also be present in `disk_attachments`.
+- `disk_attachments` (Set of String, Optional) IDs of the disks to be attached to the instance. The first entry in this set will be used to boot the instance unless `boot_disk_id` is provided. 
+- `external_ips` (Set of Object, Optional) External IP addresses attached to the instance. See [below for nested schema](#nestedatt--ips).
```
karencfv, line 89: We can't really attach/detach these, I'd probably leave it as it was

docs/resources/oxide_instance.md
```diff
@@ -64,41 +71,27 @@ resource "oxide_instance" "example" {
...
+- `network_interfaces` (Set of Object, Optional) Network interface devices attached to the instance. See [below for nested schema](#nestedatt--nics).
+- `ssh_public_keys` (Set of String, Optional) The IDs of the SSH public keys to be injected into the instance via cloud-init.
```
karencfv, line 91: "injected" sounds a little funny. What's wrong with transferred?

docs/resources/oxide_instance.md
```diff
@@ -112,11 +105,11 @@ resource "oxide_instance" "example" {
 
 ### Required
 
-- `type` (String) Type of external IP. Possible values are: ephemeral or floating.
+- `type` (String) Type of external IP. Must be one of `ephemeral` or `floating`.
 
 ### Optional
 
-- `id` (String) If type is ephemeral, ID of the IP pool to retrieve addresses from, or the current silo's default pool if not specified. If type is floating, id of the floating IP.
+- `id` (String) If `type` is `ephemeral`, the ID of the IP Pool to retrieve addresses from, or the silo's default pool if not specified. If `type` is `floating`, the ID of the floating IP.
```
karencfv, line 112:
```suggestion
- `id` (String) If `type` is `ephemeral`, the ID of the IP pool to retrieve addresses from, or the silo's default pool if not specified. If `type` is `floating`, the ID of the floating IP.
```

docs/resources/oxide_instance.md
```diff
@@ -8,50 +8,57 @@ This resource manages instances.
...
 
 ```hcl
-  project_id      = "c1dee930-a8e4-11ed-afa1-0242ac120002"
-  description     = "a test instance"
-  name            = "myinstance"
-  host_name       = "<host value>"
-  memory          = 1073741824
-  ncpus           = 1
-  start_on_create = false
```
karencfv, line 54: If this is going to be removed from this example, it should go somewhere. It'd be nice to show a use case where we're not starting the instance on creation

docs/resources/oxide_instance.md
```diff
@@ -8,50 +8,57 @@ This resource manages instances.
...
+resource "oxide_instance" "example" {
+  project_id       = "c1dee930-a8e4-11ed-afa1-0242ac120002"
+  description      = "Example instance."
+  name             = "myinstance"
+  host_name        = "myhostname"
+  memory           = 10737418240
+  ncpus            = 1
+  disk_attachments = ["611bb17d-6883-45be-b3aa-8a186fdeafe8"]
+  ssh_public_keys  = ["066cab1b-c550-4aea-8a80-8422fd3bfc40"]
+  user_data        = filebase64("path/to/init.sh")
```
karencfv, line 25: To show a minimal instance we should remove these two optional settings and add them to another more complex example.

docs/resources/oxide_instance.md
```diff
@@ -8,50 +8,57 @@ This resource manages instances.
...
   network_interfaces = [
     {
       subnet_id   = "066cab1b-c550-4aea-8a80-8422fd3bfc40"
       vpc_id      = "9b9f9be1-96bf-44ad-864a-0dedae3b3999"
-      description = "a sample nic"
+      description = "Example network interface."
       name        = "mynic"
     },
   ]
-  timeouts = {
```
karencfv, line 36: I get that these timeouts are obvious for experienced users, but I'd really like to have an example for those who are less experienced with Terraform

docs/resources/oxide_instance.md
```diff
@@ -8,50 +8,57 @@ This resource manages instances.
 
 -> Updates will stop and reboot the instance.
 
--> When setting a boot disk, the boot disk ID should also be included as part of `disk_attachments`.
```
karencfv, line 11: It's not entirely obvious that the boot disk must be part of disk attachments, and some people may not read all the way to the end. Let's leave this note please.

karencfv (commented):
Thanks for taking the time to improve the docs!

docs/resources/oxide_instance.md
sudomateo: Put the note back but modified it to use the word "must" since Terraform will error if the `boot_disk_id` is not present within `disk_attachments`.

```
Attribute "disk_attachments" must be specified when "boot_disk_id" is specified
```

docs/resources/oxide_instance.md
sudomateo: Should we move the `timeouts` documentation and example to the provider level documentation so that we're not creating a redundant example across the different resources and data sources?

docs/resources/oxide_instance.md
sudomateo: Whoops! I read it as an ordered list at first. I'll revert. It sparked another question though that I just tested. What happens when multiple disks are specified via `disk_attachments` and only one of them is bootable and it's not the first disk? The answer is that the instance will try to boot from disks in order until it finds a bootable disk or errors.

Given that context should we update this to say something along the lines of when multiple disks are provided use `boot_disk_id` to specify the boot disk?

karencfv: Yeah! That sounds good. And probably warning that otherwise the boot disk will be chosen randomly.

docs/resources/oxide_instance.md
karencfv: Hm, sure. But if we're going to remove them, they should probably all go at the same time. Otherwise, it'll be inconsistent across the docs

docs/resources/oxide_instance.md
karencfv: Excellent thanks!

karencfv (approved):
Looks great, thanks!

Issue #377, opened 2025-01-08
Support for custom router as a resource


askfongjojo:
### Overview

VPC custom router is a new type of project object that can be created, listed, and deleted. Within each router, there are also API for route addition/update/deletion. It'll be useful to provide support for router and route lifecycle management in TF.

(The priority is relatively low at this time but want to have this queued up for whenever there are cycles for small enhancement work.)

PR #423, opened 2025-04-16
Router route resource and datasource

karencfv:
Tested against the dogfood rack:

```console
$ TEST_ACC_NAME=TestAccCloudResourceVPCRouterRoute_full make testacc
=== RUN   TestAccCloudResourceVPCRouterRoute_full
=== PAUSE TestAccCloudResourceVPCRouterRoute_full
=== CONT  TestAccCloudResourceVPCRouterRoute_full
--- PASS: TestAccCloudResourceVPCRouterRoute_full (17.90s)
PASS
ok  	github.com/oxidecomputer/terraform-provider-oxide/internal/provider	18.327s
$ TEST_ACC_NAME=TestAccCloudDataSourceVPCRouterRoute_full make testacc
=== RUN   TestAccCloudDataSourceVPCRouterRoute_full
=== PAUSE TestAccCloudDataSourceVPCRouterRoute_full
=== CONT  TestAccCloudDataSourceVPCRouterRoute_full
--- PASS: TestAccCloudDataSourceVPCRouterRoute_full (1.87s)
PASS
ok  	github.com/oxidecomputer/terraform-provider-oxide/internal/provider	2.148s
```

Closes: https://github.com/oxidecomputer/terraform-provider-oxide/issues/377


karencfv:
Blocked on https://github.com/oxidecomputer/omicron/issues/7996

karencfv:
Will be unblocked by https://github.com/oxidecomputer/oxide.go/pull/284

karencfv:
tiny ping @sudomateo do you mind taking a look?

internal/provider/data_source_vpc_router_route.go
```diff
@@ -0,0 +1,241 @@
...
+	attributeTypes := map[string]attr.Type{
+		"type":  types.StringType,
+		"value": types.StringType,
+	}
+	destination, diags := types.ObjectValueFrom(ctx, attributeTypes, dm)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	state.Destination = destination
```
sudomateo, line 205: I'm curious why you went with `types.Object` for `Destination` instead of a pointer to a `vpcRouterRouteDestinationDataSourceModel` like in the `oxide_image` data source. 

https://github.com/oxidecomputer/terraform-provider-oxide/blob/9f0e000cda86ea2c19b3b385c6f48acea84a95d8/internal/provider/data_source_image.go#L40

Then you could have skipped the parsing here. I'm not familiar with which way is technically superior but I figured I'd ask you about it.

internal/provider/data_source_vpc_router_route.go
```diff
@@ -0,0 +1,241 @@
...
+	targetAttributeTypes := map[string]attr.Type{
+		"type":  types.StringType,
+		"value": types.StringType,
+	}
+	target, diags := types.ObjectValueFrom(ctx, targetAttributeTypes, tm)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	state.Target = target
```
sudomateo, line 226: Same why `types.Object` comment from earlier about the `Target` field.

internal/provider/resource_vpc_router_route.go
```diff
@@ -0,0 +1,408 @@
...
+			Name: oxide.Name(plan.Name.ValueString()),
+			Target: oxide.RouteTarget{
+				Type: oxide.RouteTargetType(plan.Target.Type.ValueString()),
+			},
+		},
+	}
+
+	if !plan.Target.Value.IsNull() {
+		params.Body.Target.Value = plan.Target.Value.ValueString()
+	}
```
sudomateo, line 215: `ValueString` already returns an empty string if its value is null or unknown in the plan. We shouldn't need this check due to that behavior. Did you run into any behavior that led to creating this logic?

internal/provider/resource_vpc_router_route.go
```diff
@@ -0,0 +1,408 @@
...
+			},
+			Target: oxide.RouteTarget{
+				Type: oxide.RouteTargetType(plan.Target.Type.ValueString()),
+			},
+		},
+	}
+
+	if !plan.Target.Value.IsNull() {
+		params.Body.Target.Value = plan.Target.Value.ValueString()
+	}
```
sudomateo, line 351: Same comment from `Create`.

sudomateo (approved):
Changes look good overall. I left a small typo suggestion for documentation and asked a few non-blocking questions. Feel free to move forward with this on your own timeline. Thank you for implementing this!

internal/provider/data_source_vpc_router_route.go
karencfv: Frankly, while developing I got an error suggesting I use `types.Object` so I went with that

internal/provider/data_source_vpc_router_route.go
karencfv: same

internal/provider/resource_vpc_router_route.go
karencfv: Ah, no! I was trying out a different type in my attempt to deal with enums and forgot to remove that. Thanks!

karencfv: Actually no, checked it again. `Body.Target.Value` is type any, not string. So we don't want to set it as an empty string accidentally. I'll add a comment

karencfv:
Thanks for the review!!

PR #407, opened 2025-03-15
Silo Creation feature

gurumachupalli:

PR #425, opened 2025-04-17
Feature silo creation

gurumachupalli:


sudomateo:
This is picked up from #407. I'll get to reviewing this later today or, at worse, tomorrow.

gurumachupalli:
> This is picked up from #407. I'll get to reviewing this later today or, at worse, tomorrow.

Thanks Matthew!

sudomateo (changes requested):
I cloned your branch locally and did some testing. I have most of these changes already done locally but I'll finish them up tomorrow and push them. Here's what I found that needed some work.

* [x] The `Resources` method in `internal/provider/provider.go` file needs to be updated to call the `NewSiloResource` constructor.
* [x] The `Cpus` field takes a `*int` which can be provided by calling `oxide.NewPointer`.
* [x] The `Read` method needs to be updated to make an API call to [silo_quotas_view](https://docs.oxide.computer/api/silo_quotas_view) to read the latest state of quotas. This will allow Terraform to detect quotas drift.
* [x] The `mapped_fleet_roles` attribute needs to be set to `nil` when unset in the configuration. Otherwise Terraform will detect drift since the `to*MappedFleetRoles` functions return an empty map rather than `nil`.
* [x] The `Update` method needs to call [silo_view](https://docs.oxide.computer/api/silo_view) after updating the silo quotas to fetch the `time_created` and `time_modified` values. Really just the `time_modified` value since that will change. Without doing this then Terraform will produce an inconsistent plan.
* [x] The `Memory` and `Storage` fields are `uint64` but Terraform can only express `int64` types. We'll have to add a validator that checks if their values are at least 0 and then cast the `int64` to a `uint64` in the places that read/write those fields.

sudomateo:
Okay I'm liking where this is now. I'll just take a final look tomorrow at the code, update the docs, update the tests, and then this will be ready for merging.

sudomateo:
I removed the import functionality from this resource since it was resulting in a replacement due to the `tls_certificates` attribute. I filed https://github.com/oxidecomputer/terraform-provider-oxide/issues/430 to track fixing this but in the interest of time I'll move forward with the current functionality.

sudomateo (approved):

sudomateo:
@karencfv since I pushed commits to this branch would you mind giving this a review?

sudomateo:
> I removed the import functionality from this resource since it was resulting in a replacement due to the `tls_certificates` attribute. I filed #430 to track fixing this but in the interest of time I'll move forward with the current implementation that does not have import functionality.

I added import functionality in https://github.com/oxidecomputer/terraform-provider-oxide/pull/425/commits/4a6c0760a9375de5632cb39ea852ebd79d5007ae after learning about [write-only attributes](https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments). However, it requires Terraform v1.11 which I noted in the documentation.

docs/resources/oxide_silo.md
```diff
@@ -0,0 +1,100 @@
+---
+page_title: "oxide_silo Resource - terraform-provider-oxide"
+---
+
+# oxide_silo (Resource)
+
+This resource manages the creation of an Oxide silo.
+
+-> This resource requires Terraform version v1.11 or later due to the
```
karencfv, line 9: Minimum Terraform versions shouldn't be per resource, but rather globally. Otherwise, it can get pretty confusing for users to know which resources require which version.

I suggest we change the minimum required version in the README and in docs/index.md . Additionally, this minimum version required should be noted as a breaking change in the changelog. We can note that the only resource that requires it is `oxide_silo`

I know updating versions will be annoying for customers. But, better bite the bullet sooner rather than later.

internal/provider/resource_silo.go
```diff
@@ -0,0 +1,551 @@
...
+				"API error: "+err.Error(),
+			)
+			return
+		}
+	}
+
+	tflog.Trace(ctx, fmt.Sprintf("deleted silo with ID: %v", state.ID.ValueString()), map[string]any{"success": true})
+}
+
+func toOxideMappedFleetRoles(mappedFleetRoles map[string][]string) map[string][]oxide.FleetRole {
```
karencfv, line 510: Nit: 
```suggestion
func stringMapToFleetRoleMap(mappedFleetRoles map[string][]string) map[string][]oxide.FleetRole {
```

internal/provider/resource_silo.go
```diff
@@ -0,0 +1,551 @@
...
+		var roles []oxide.FleetRole
+		for _, frm := range fleetRoleModels {
+			roles = append(roles, oxide.FleetRole(frm))
+		}
+		model[key] = roles
+	}
+	return model
+}
+
+func toOxideTlsCertificates(tlsCertificates []siloResourceTlsCertificateModel) []oxide.CertificateCreate {
```
karencfv, line 523: Nit:
```suggestion
func tlsCertsModelToCertificateCreateSlice(tlsCertificates []siloResourceTlsCertificateModel) []oxide.CertificateCreate {
```

docs/resources/oxide_silo.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+## Schema
+
+### Required
+
+- `name` (String) Name of the Oxide silo.
+- `description` (String) Description for the Oxide silo.
+- `quotas` (Set of Object) Limits the amount of provisionable CPU, memory, and storage in the silo. (See [below for nested schema](#nestedatt--quotas).)
+- `tls_certificates` (String, Write-only) Initial TLS certificates to be used for the new silo's console and API endpoints. This attribute is a [write-only attribute](https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments) and can only be modified by updating its configuration and [replacing the resource](https://developer.hashicorp.com/terraform/cli/state/taint). (See [below for nested schema](#nestedatt--tls).)
```
karencfv, line 53: It appears we're not requiring a replace on the schema https://github.com/oxidecomputer/terraform-provider-oxide/pull/425/files#diff-1ea56842b20e164d37586563e1ee70d692dbc252771685f2072ef22bc4706682R153-R160.

Also, perhaps we don't need to specify that this attribute requires replace? There are other attributes that require replace and we don't mention it in the docs (e.g. `name`).

docs/resources/oxide_silo.md
```diff
@@ -0,0 +1,100 @@
...
+- `discoverable` (Boolean) Whether this silo is present in the silo_list output. Defaults to `true`.
+
+### Optional
+
+- `identity_mode` (String) How identities are managed and users are authenticated in this silo. Valid values are `saml_jit` and `local_only`. Defaults to `local_only`.
+- `admin_group_name` (String) This group will be created during silo creation and granted the "Silo Admin" role. Identity providers can assert that users belong to this group and those users can log in and further initialize the Silo.
+- `mapped_fleet_roles` (Map) Mapping of which fleet roles are conferred by each silo role. The default is that no fleet roles are conferred by any silo roles unless there's a corresponding entry in this map. Map keys must be one of `admin`, `collaborator`, or `viewer` and map values are an array of strings that must be one of `admin`, `collaborator`, or `viewer`.
```
karencfv, line 60: This paragraph is a little confusing. Could you rephrase?

docs/resources/oxide_silo.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Nested Schema for `tls_certificates`
+
+### Required
+
+- `name` (String) The name associated with the certificate.
+- `description` (String) Description of the certificate.
+- `cert` (String) PEM-formatted string containing public certificate chain.
+- `key` (String) PEM-formatted string containing private key.
+- `service` (String) The service associated with the certificate. The only valid value is `external_api`. Defaults to `external_api`.
```
karencfv, line 89: Should we even expose this attribute if the only option is `external_api`? We can make this an optional field when there are other choices

internal/provider/resource_silo.go
```diff
@@ -0,0 +1,551 @@
...
+			},
+			"name": schema.StringAttribute{
+				Required:    true,
+				Description: "Unique, immutable, user-controlled identifier of the silo.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+				Validators: []validator.String{
+					stringvalidator.RegexMatches(
+						regexp.MustCompile(`^[a-zA-Z0-9-]+$`),
```
karencfv, line 122: Let's make this value a constant so it can be reused in other resources

internal/provider/resource_silo.go
```diff
@@ -0,0 +1,551 @@
...
+					},
+				},
+			},
+			"tls_certificates": schema.ListNestedAttribute{
+				Required:    true,
+				WriteOnly:   true,
+				Description: "Initial TLS certificates to be used for the new silo's console and API endpoints.",
+				Validators: []validator.List{
+					listvalidator.SizeAtLeast(1),
+				},
```
karencfv, line 159: Is this necessary? If an attribute is required, Terraform should check that it has at least one item no?

internal/provider/resource_silo.go
```diff
@@ -0,0 +1,551 @@
...
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"name": schema.StringAttribute{
+							Required:    true,
+							WriteOnly:   true,
+							Description: "Unique, immutable, user-controlled identifier of the certificate.",
+							Validators: []validator.String{
+								stringvalidator.RegexMatches(
+									regexp.MustCompile(`^[a-zA-Z0-9-]+$`),
```
karencfv, line 168: Same about the constant

internal/provider/resource_silo.go
```diff
@@ -0,0 +1,551 @@
...
+			AdminGroupName:   plan.AdminGroupName.ValueString(),
+			Description:      plan.Description.ValueString(),
+			IdentityMode:     oxide.SiloIdentityMode(plan.IdentityMode.ValueString()),
+			Discoverable:     plan.Discoverable.ValueBoolPointer(),
+			MappedFleetRoles: toOxideMappedFleetRoles(plan.MappedFleetRoles),
+			Name:             oxide.Name(plan.Name.ValueString()),
+			Quotas: oxide.SiloQuotasCreate{
+				Cpus:    oxide.NewPointer(int(plan.Quotas.Cpus.ValueInt64())),
+				Memory:  oxide.ByteCount(plan.Quotas.Memory.ValueInt64()),
+				Storage: oxide.ByteCount(plan.Quotas.Storage.ValueInt64()),
```
karencfv, line 306: Let's use `ValueInt64Pointer` instead of `ValueInt64`

internal/provider/resource_silo.go
```diff
@@ -0,0 +1,551 @@
...
+		return
+	}
+
+	state.ID = types.StringValue(silo.Id)
+	state.Name = types.StringValue(string(silo.Name))
+	state.Description = types.StringValue(silo.Description)
+	state.Quotas = &siloResourceQuotasModel{
+		Cpus:    types.Int64Value(int64(*siloQuotas.Cpus)),
+		Memory:  types.Int64Value(int64(siloQuotas.Memory)),
+		Storage: types.Int64Value(int64(siloQuotas.Storage)),
```
karencfv, line 389: Same here, we should use `Int64PointerValue`, no?

internal/provider/resource_silo.go
```diff
@@ -0,0 +1,551 @@
...
+
+	ctx, cancel := context.WithTimeout(ctx, updateTimeout)
+	defer cancel()
+
+	siloQuotasParams := oxide.SiloQuotasUpdateParams{
+		Silo: oxide.NameOrId(state.ID.ValueString()),
+		Body: &oxide.SiloQuotasUpdate{
+			Cpus:    oxide.NewPointer(int(plan.Quotas.Cpus.ValueInt64())),
+			Memory:  oxide.ByteCount(plan.Quotas.Memory.ValueInt64()),
+			Storage: oxide.ByteCount(plan.Quotas.Storage.ValueInt64()),
```
karencfv, line 434: Same as above

internal/provider/resource_silo.go
```diff
@@ -0,0 +1,551 @@
...
+			"API error: "+err.Error(),
+		)
+		return
+	}
+
+	plan.ID = types.StringValue(siloQuotas.SiloId)
+	plan.Quotas = &siloResourceQuotasModel{
+		Cpus:    types.Int64Value(int64(*siloQuotas.Cpus)),
+		Memory:  types.Int64Value(int64(siloQuotas.Memory)),
+		Storage: types.Int64Value(int64(siloQuotas.Storage)),
```
karencfv, line 464: Same as above

internal/provider/resource_silo.go
```diff
@@ -0,0 +1,551 @@
...
+			Service:     oxide.ServiceUsingCertificate(tlsCert.Service.ValueString()),
+		}
+
+		model = append(model, r)
+	}
+
+	return model
+}
+
+func toTerraformMappedFleetRoles(mappedFleetRoles map[string][]oxide.FleetRole) map[string][]string {
```
karencfv, line 541:
```suggestion
func fleetRoleMapToStringMap(mappedFleetRoles map[string][]oxide.FleetRole) map[string][]string {
```

internal/provider/resource_silo_test.go
```diff
@@ -0,0 +1,219 @@
...
...DVlb28KQWlHSFB3OC92TnpHd2hoaXZvSExBZ01CQUFHalV6QlJNQjBHQTFVZERnUVdCQlJQRlhUdm83OHdwUWtLaHhDMgo0OXpnL01DSTZEQWZCZ05WSFNNRUdEQVdnQlJQRlhUdm83OHdwUWtLaHhDMjQ5emcvTUNJNkRBUEJnTlZIUk1CCkFmOEVCVEFEQVFIL01BMEdDU3FHU0liM0RRRUJDd1VBQTRJQ0FRQ0VmRFpzZTRIRW1PME03dWRjR2hHeWZpTXoKN3A2Y0lXQS9XVThaSEFLT04zSGlwYUpwalp4OE1aVkVaMzVobTRROFEzYkxoRHpZVTJmV3RMMXhBN0JyaHZCMQpudGN0VXRVOGtCN1JKbzl3VjU5UFREaFhzMzRiV1JyZ1lzZkRFL3ExZjNCQ2xRanhBTTA0T3hJeXVKSzNkendUCnBvbXdNaFpKSmRYcXhoL1NnR0IwdzhZUkw5WEV6dVh1aFJoNmk2b0N2Y0FQQ0VDcjVuQllraWlVSmJiams4YVMKdDVQMVJSTFdEV2MrSGloYWx4cmlFRjRUbFFUbW96N2cxdzZJVDdYZjFkZ1A1aTlaN1Vad1RWQlRnZG1naW5JVgp4SnFTOVJNN3pVSTJqWHU5aTk0Zk5qOVA4VUJWRysxS3BqazQvaE5VOWp2TGNGRFNPRjZkSnZwRmFmKzg1di9WCjJ2WFVMb1FJV3IvcDF2dUdMTmxnb0M1WFptN2ZzdDlRcEdva1hlMEwySWp0Q1pncWRKRzVBOHhiaDJEa1orbEQKNXk5akVMOTRGRXNBTndFelhxQlVDQWpjZ2pwUVJBajM2Nk9tb1pSVDFCeUNYK0RXNFBXRnJvclorc0Z6VStmNQpCQUpFaTJVNmg2VGdaQ2d1Ty81Z0l6QWtDL0xXdThYb0lwNExGVnVWRmhvQzJSZm40YWhxdm42RHNSdTVqcWoyCnhIa1RLUVdmZlN3dGQ5akRQWm9Db2JEYkFXREdOWllTQ2tyMU5icmZYOGRBL3cwK2ZXb1NMVjhiOEhzMG5CSUgKNWpkdkJOQTdVR1UzK0tnamFrenJSRVkwWnRJUEJGdEZZVGMxb3Y0aHQrZVpxeEQxcEh5VFArMHltYkJYZXlaeQpKMlVJRnQwUThQSzhOUkdXRkE9PQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg=="
```
karencfv, line 23: What happens if we truncate these? Does the API still accept them?

If the answer is yes then let's do that. We are not testing a fully functioning silo here, just that we are writing to the API correctly.

karencfv (commented):
Thanks for working on this! looks like a tricky one to figure out 😅

internal/provider/resource_silo_test.go
```diff
@@ -0,0 +1,219 @@
...
+      description = "Self-signed wildcard certificate for *.sys.r3.oxide-preview.com."
+      cert        = base64decode("{{.TLSCertificateBase64}}")
+      key         = base64decode("{{.TLSPrivateKeyBase64}}")
+      service     = "external_api"
+    },
+  ]
+}
+`
+
+func TestAccCloudResourceSilo_full(t *testing.T) {
```
karencfv, line 101: The prefix `TestAccCloud` is reserved for "cloud" resources that we run as a on every release against the colo rack. 

```suggestion
func TestAccSiloResourceSilo_full(t *testing.T) {
```

karencfv:
@sudomateo do you mind running the acceptance tests against a dev environment and pasting the results here, as we can't run these as part of CI yet?

docs/resources/oxide_silo.md
gurumachupalli: @karencfv So confirming that I should remove move this note to the README.md and in the changelog state that it's a breaking change?

karencfv: Yeah, that'd be great thanks!

internal/provider/resource_silo.go
gurumachupalli: my Go knowledge might be lacking. trying to understand why it should be `Int64PointerValue`, i have this for my `siloResourceQuotasModel` struct.

`type siloResourceQuotasModel struct {
	Cpus    types.Int64 `tfsdk:"cpus"`
	Memory  types.Int64 `tfsdk:"memory"`
	Storage types.Int64 `tfsdk:"storage"`
}`

The base type that I only saw is a `Int64Value` (https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework@v1.14.1/types#Int64). So wouldn't this stay as is?

internal/provider/resource_silo_test.go
gurumachupalli: that i'm not sure. i am having trouble running the acceptance tests to test this, would you or Matthew be able to test this? I tried running `make testacc` but was running into problems, i had set the test environment variables as well. 

`NAME  TestAccCloudResourceInstance_extIPs
    resource_instance_test.go:242: Step 1/2 error: Error running pre-apply plan: exit status 1
        
        Error: Failed to load plugin schemas
        
        Error while loading schemas for plugin components: Failed to obtain provider
        schema: Could not load the schema for provider
        registry.terraform.io/hashicorp/oxide: failed to retrieve schema from
        provider "registry.terraform.io/hashicorp/oxide": Schema Using Attribute
        Default For Non-Computed Attribute: Attribute "tls_certificates.service" must
        be computed when using default. This is an issue with the provider and should
        be reported to the provider developers...`
        
An example of the failures I was seeing ^

internal/provider/resource_silo.go
gurumachupalli: Thought it'd be good to have an extra check. But want me to remove?

internal/provider/resource_silo.go
karencfv: Ah so sorry! This comment was meant for the code two lines above `Cpus: types.Int64Value(int64(*siloQuotas.Cpus)),`.

We want to guard for the case of siloQuotas.Cpus being nil

internal/provider/resource_silo_test.go
karencfv: > Schema Using Attribute Default For Non-Computed Attribute: Attribute "tls_certificates.service" must be computed when using default.

Aha! it seems like is a bug in the schema code. It appears you cannot have a `Default` for an attribute that is not computed. `tls_certificates.service` https://github.com/oxidecomputer/terraform-provider-oxide/pull/425/files#diff-1ea56842b20e164d37586563e1ee70d692dbc252771685f2072ef22bc4706682R191 should be computed as well to be able to use defaults.

karencfv: Oh! Btw, to run this test only, instead of the whole suite, you'll want to do

```
$ TEST_ACC_NAME=TestAccSiloResourceSilo_full make testacc
```

docs/resources/oxide_silo.md
sudomateo: We can't require a replace on this since it's a write-only attribute and the API does not return the `tls_certificates` back to us. If we required a replace we'd always have a diff since the configuration would not match what's returned from `Read`.

docs/resources/oxide_silo.md
sudomateo: I was a little confused on this myself when writing it. The API documentation is a bit lacking here so I tested this by hitting the API directly and reading the error messages that read that only `admin`, `collaborator`, and `viewer` can be valid keys and valid values in the array.

docs/resources/oxide_silo.md
karencfv: Gotcha, could we rephrase `and can only be modified by updating its configuration and [replacing the resource](https://developer.hashicorp.com/terraform/cli/state/taint).` then? I think I was a bit confused by this sentence in particular.

docs/resources/oxide_silo.md
sudomateo: I'm okay with making it `Optional` with a `Default` and a validator. I don't really have a preference honestly.

sudomateo:
> @sudomateo do you mind running the acceptance tests against a dev environment and pasting the results here, as we can't run these as part of CI yet?

This was run from commit https://github.com/oxidecomputer/terraform-provider-oxide/pull/425/commits/4a6c0760a9375de5632cb39ea852ebd79d5007ae.

```
> TEST_ACC_NAME=TestAccCloudResourceSilo_full make testacc
=== RUN   TestAccCloudResourceSilo_full
=== PAUSE TestAccCloudResourceSilo_full
=== CONT  TestAccCloudResourceSilo_full
--- PASS: TestAccCloudResourceSilo_full (4.50s)
PASS
ok      github.com/oxidecomputer/terraform-provider-oxide/internal/provider     4.509s
```

docs/resources/oxide_silo.md
karencfv: Thanks! I'm guessing we're assuming the users know what they're doing here? Do we have a link to some docs explaining how this works? For example, what would something like `"viewer"={admin, collaborator}` mean?

internal/provider/resource_silo_test.go
sudomateo: From the [write-only arguments documentation](https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments).

> If a list nested, map nested, or single nested attribute has WriteOnly set to true, all child attributes must also have WriteOnly set to true.
>
> Computed cannot be set to true for write-only arguments.

We can't have `service` set to `Computed` since it's within the write-only `tls_certificates` attribute.

sudomateo: > What happens if we truncate these? Does the API still accept them?

Sorry I missed this question at first. Unfortunately the API not only checks that the TLS certificate is a valid PEM-encoded certificate but it also verifies whether the Common Name (CN) is valid for the silo name that's attempting to be created.

karencfv: Oh well, was worth trying out 😅

sudomateo:
It's quite late here but please do ping if you need me. I can take a further look at this tomorrow and help get it over the line and implement whatever feedback is remaining.

karencfv:
> It's quite late here but please do ping if you need me. I can take a further look at this tomorrow and help get it over the line and implement whatever feedback is remaining.

Go rest! this can wait for tomorrow :)

internal/provider/resource_silo.go
gurumachupalli: gotcha makes sense! pushed the fix @karencfv

docs/resources/oxide_silo.md
gurumachupalli: made the change to optional and updated the docs

internal/provider/resource_silo.go
```diff
@@ -0,0 +1,568 @@
...
+
+	ctx, cancel := context.WithTimeout(ctx, updateTimeout)
+	defer cancel()
+
+	siloQuotasParams := oxide.SiloQuotasUpdateParams{
+		Silo: oxide.NameOrId(state.ID.ValueString()),
+		Body: &oxide.SiloQuotasUpdate{
+			Cpus:    oxide.NewPointer(int(*plan.Quotas.Cpus.ValueInt64Pointer())),
+			Memory:  oxide.ByteCount(*plan.Quotas.Memory.ValueInt64Pointer()),
+			Storage: oxide.ByteCount(*plan.Quotas.Storage.ValueInt64Pointer()),
```
karencfv, line 440: I think you can just use `ValueInt64` here. It seems you're setting a pointer and then dereferencing it immediately

karencfv (commented):
Thanks for taking the time to address my concerns! Just leaving a few more comments

internal/provider/resource_silo.go
gurumachupalli: the previous one i had was `Int64Value` so changing it back to fit the model

gurumachupalli:
@karencfv made the fixes you suggested! thanks for the feedback

docs/resources/oxide_silo.md
sudomateo: This was updated via another discussion on this pull request. Resolving.

internal/provider/resource_silo.go
```diff
@@ -0,0 +1,568 @@
...
+							WriteOnly:   true,
+							Description: "PEM-formatted string containing private key.",
+						},
+						"service": schema.StringAttribute{
+							Optional:    true,
+							WriteOnly:   true,
+							Description: "Service using this certificate.",
+							Default:     stringdefault.StaticString("external_api"),
+							Validators: []validator.String{
+								stringvalidator.OneOf("external_api"),
```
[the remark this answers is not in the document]
sudomateo: We can't do this unfortunately. From the [documentation](https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments).

> If a list nested, map nested, or single nested attribute has `WriteOnly` set to `true`, all child attributes must also have `WriteOnly` set to `true`.
>
> `Computed` cannot be set to true for write-only arguments.

Here's what the acceptance tests return when we change `Optional` to `Computed`.

```
> TEST_ACC_NAME=TestAccSiloResourceSilo_full make testacc
=== RUN   TestAccSiloResourceSilo_full
=== PAUSE TestAccSiloResourceSilo_full
=== CONT  TestAccSiloResourceSilo_full
    resource_silo_test.go:131: Step 1/3 error: Error running pre-apply plan: exit status 1

        Error: Invalid Attribute Definition

          with oxide_silo.acc-silo-a9c0d466-2382-426f-aad9-caa14c2bd29a,
          on terraform_plugin_test.tf line 12, in resource "oxide_silo" "acc-silo-a9c0d466-2382-426f-aad9-caa14c2bd29a":
          12: resource "oxide_silo" "acc-silo-a9c0d466-2382-426f-aad9-caa14c2bd29a" {

        WriteOnly Attributes cannot be set with Computed. This is always a problem
        with the provider and should be reported to the provider developer.
--- FAIL: TestAccSiloResourceSilo_full (0.28s)
FAIL
FAIL    github.com/oxidecomputer/terraform-provider-oxide/internal/provider     0.282s
FAIL
make: *** [Makefile:82: testacc] Error 1
```

Here's what they look like if we leave it `Optional` with a `Default`.

```
> TEST_ACC_NAME=TestAccSiloResourceSilo_full make testacc
=== RUN   TestAccSiloResourceSilo_full
=== PAUSE TestAccSiloResourceSilo_full
=== CONT  TestAccSiloResourceSilo_full
    resource_silo_test.go:131: Step 1/3 error: Error running pre-apply plan: exit status 1

        Error: Failed to load plugin schemas

        Error while loading schemas for plugin components: Failed to obtain provider
        schema: Could not load the schema for provider
        registry.terraform.io/hashicorp/oxide: failed to retrieve schema from
        provider "registry.terraform.io/hashicorp/oxide": Schema Using Attribute
        Default For Non-Computed Attribute: Attribute "tls_certificates.service" must
        be computed when using default. This is an issue with the provider and should
        be reported to the provider developers...
--- FAIL: TestAccSiloResourceSilo_full (0.08s)
FAIL
FAIL    github.com/oxidecomputer/terraform-provider-oxide/internal/provider     0.082s
FAIL
make: *** [Makefile:82: testacc] Error 1
```

We have the following options. Let me know if I'm missing anything else.

* Remove `tls_certificates.service` from the schema entirely and hard code it in the underlying API call.
* Set it to `Required`, remove `Default`, and validate its value.
* Don't use `WriteOnly` on the entire `tls_certificates` attribute. This is really a non-option since then we'd have drift every plan/apply.

I've changed it to `Required` for now and the tests pass. Happy to change it again if we decide it should be different.

```
> TEST_ACC_NAME=TestAccSiloResourceSilo_full make testacc
=== RUN   TestAccSiloResourceSilo_full
=== PAUSE TestAccSiloResourceSilo_full
=== CONT  TestAccSiloResourceSilo_full
--- PASS: TestAccSiloResourceSilo_full (3.49s)
PASS
ok      github.com/oxidecomputer/terraform-provider-oxide/internal/provider     3.495s
```

karencfv (approved):
Looks great! Thanks for bearing with me on this!

internal/provider/resource_silo.go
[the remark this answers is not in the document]
karencfv: Ahhhh that's a bummer. Thanks for looking into this though!

PR #442, opened 2025-05-30
oxide_silo_saml_identity_provider: initial resource

sudomateo:
Added the `oxide_silo_saml_identity_provider` resource. Note that this resource does not support updates or deletion.

### DIFF
--- a/.changelog/0.14.0.toml
+++ b/.changelog/0.14.0.toml
@@ -3,6 +3,8 @@ title = ""
 description = ""
 
 [[features]]
+title = "New resource"
+description = "`oxide_silo_saml_identity_provider` [#442](https://github.com/oxidecomputer/terraform-provider-oxide/pull/442)."
 
 [[enhancements]]
 title = ""
--- a/docs/resources/oxide_silo_saml_identity_provider.md
+++ b/docs/resources/oxide_silo_saml_identity_provider.md
@@ -1,0 +1,115 @@
+---
+page_title: "oxide_silo_saml_identity_provider Resource - terraform-provider-oxide"
+---
+
+# oxide_silo_saml_identity_provider (Resource)
+
+Manages a SAML identity provider (IdP) for an Oxide silo.
+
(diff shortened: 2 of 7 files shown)


sudomateo:
> We'll be releasing R15 soon, and all of our customer's racks are still updated by us. So I don't think there should be much of a problem waiting for this feature a few weeks more right?
> 
> If we add the functionality to reference names, it will be much more problematic to remove it in the future.

I agree. I'll ask the customer who asked for this feature to see what their urgency and timelines are. I don't see why it can't wait for R15 especially since it'll be the better implementation.

sudomateo:
There's no upstream update or delete API for this resource, so the following quirks should be understood.

I omitted import support for this resource since the read endpoint does not return `silo` or `idp_metadata_source` and a configuration that contains these attributes would attempt the import then immediately try to perform an `Update` which is not supported.

I also chose to add a warning diagnostic instead of an error diagnostic in `Delete` so that Terraform will remove the resource from state but not from Oxide since there's no delete API. This allows users to run `terraform destroy` to destroy both the `oxide_silo_saml_identity_provider` and `oxide_silo` resources when testing IdP configurations.

I'll add tests, documentation, and a changelog entry next week.

sudomateo:
Tests look good.

```
> TEST_ACC_NAME=TestAccSiloResourceSiloSamlIdentityProvider OXIDE_SILO_DNS_NAME='*.sys.r3.oxide-preview.com' make testacc
=== RUN   TestAccSiloResourceSiloSamlIdentityProvider
=== PAUSE TestAccSiloResourceSiloSamlIdentityProvider
=== CONT  TestAccSiloResourceSiloSamlIdentityProvider
--- PASS: TestAccSiloResourceSiloSamlIdentityProvider (5.17s)
PASS
ok      github.com/oxidecomputer/terraform-provider-oxide/internal/provider     5.185s
```

lgfa29 (approved):

Issue #466, opened 2025-07-14
`oxide_instance`: provider crash when `boot_disk_id` is empty


sudomateo:
The following configuration causes a provider crash.

```hcl
resource "oxide_instance" "example" {
  project_id       = data.oxide_project.example.id
  description      = "example"
  name             = "example"
  host_name        = "example"
  memory           = 1024 * 1024 * 1024
  ncpus            = 2
  disk_attachments = [oxide_disk.example.id]
}
```

Here's the crash.

```
oxide_instance.example: Creating...
╷
│ Error: Request cancelled
│
│ The plugin6.(*GRPCProvider).ApplyResourceChange request was cancelled.
╵

Stack trace from the terraform-provider-oxide_v0.12.0 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xbe1e93]

goroutine 19 [running]:
github.com/oxidecomputer/terraform-provider-oxide/internal/provider.filterBootDiskFromDisks(...)
        github.com/oxidecomputer/terraform-provider-oxide/internal/provider/resource_instance.go:1221
github.com/oxidecomputer/terraform-provider-oxide/internal/provider.(*instanceResource).Create(0xc00043aff8, {0x101c4b8, 0xc000172570}, {{{{0x1021b18, 0xc00017a4b0}, {0xd8fc60, 0xc000173da0}}, {0x1023920, 0xc000416370}}, {{{0x1021b18, ...}, ...}, ...}, ...}, ...)
        github.com/oxidecomputer/terraform-provider-oxide/internal/provider/resource_instance.go:417 +0xc73
github.com/hashicorp/terraform-plugin-framework/internal/fwserver.(*Server).CreateResource(0xc000186a08, {0x101c4b8, 0xc000172570}, 0xc0001274c0, 0xc000127490)
        github.com/hashicorp/terraform-plugin-framework@v1.15.0/internal/fwserver/server_createresource.go:127 +0x88e
github.com/hashicorp/terraform-plugin-framework/internal/fwserver.(*Server).ApplyResourceChange(0xc000186a08, {0x101c4b8, 0xc000172570}, 0xc00041e310, 0xc0001275e8)
        github.com/hashicorp/terraform-plugin-framework@v1.15.0/internal/fwserver/server_applyresourcechange.go:63 +0x586
[...]

PR #471, opened 2025-07-17
resource(instance): fix provider crash when boot_disk_id is empty

sudomateo:
Added a nil pointer check to prevent a nil pointer dereference causing the provider to crash when `boot_disk_id` is empty.

The following configuration now successfully applies.

```hcl
  project_id       = data.oxide_project.example.id
  description      = "example"
  name             = "example"
  host_name        = "example"
  memory           = 1024 * 1024 * 1024
  ncpus            = 2
  disk_attachments = [oxide_disk.example.id]
}
```

Fixes https://github.com/oxidecomputer/terraform-provider-oxide/issues/466.


karencfv (commented):
Thanks for fixing this! I'll wait for the acceptance test to review again :)

karencfv:
Looks like there's already a test for this case

https://github.com/oxidecomputer/terraform-provider-oxide/blob/7af9fa6dca7deb768f1ff692d686a801780070b7/internal/provider/resource_instance_test.go#L46-L55

I'm surprised this only just started failing! Must have been some change in the API we didn't catch. I guess just running the existing instance tests should be fine to test this change then!

Issue #476, opened 2025-08-01
Support SNAT IPs in external IP list


sudomateo:
The `instance_external_ip_list` API was updated to return SNAT IPs as well as floating IPs and ephemeral IPs in https://github.com/oxidecomputer/omicron/pull/8685. Update the Terraform provider accordingly.

Needs https://github.com/oxidecomputer/oxide.go/issues/306 as well.

PR #477, opened 2025-08-01
provider: support snat ips in external ips listing

sudomateo:
Updated the Terraform provider to support SNAT IPs being returned from the `instance_external_ip_list` API.

Needs the change in https://github.com/oxidecomputer/oxide.go/issues/306.

Closes https://github.com/oxidecomputer/terraform-provider-oxide/issues/476


karencfv (commented):
Looks good! Just wondering why we're not introducing support for attaching/detaching SNAT IPs as well?

karencfv (approved):
> There's no support for attaching an SNAT IP in Omicron, just reading them

Huh 🤔 that's odd! Oh well, looks great then!

PR #478, opened 2025-08-02
resource(instance): use floating ip id instead of ip pool id

sudomateo:
Updated the `oxide_instance` resource to use the floating IP ID for floating external IPs instead of the IP pool ID. Without this change, Terraform will be unable to detach a floating IP from an instance since it will try to detach using the floating IP IP pool ID instead of the floating IP ID itself.


sudomateo:
Found this while testing some Kubernetes deployments. Will look at it further next week.

PR #487, opened 2025-08-08
Update codeowners file

karencfv:
Now that there is an official solutions SW engineering team I will be handing over the reigns. I'll still be around for reviews, questions, etc if anyone needs them. See you all around! 👋


sudomateo (approved):
You're always welcome here! Thanks for all your help onboarding me.

PR #494, opened 2025-08-18
misc: update tests to pass against omicron dev.

jmcarp:
Update tests and docs such that the entire acceptance test suite runs against the simulated omicron environment. We either document and script all resources that the tests depend on, or teach the tests to create the dependencies themselves. We'll look into running the acceptance suite against simulated omicron in a separate patch.

PR #497, opened 2025-08-19
misc: run acceptance tests from github actions.

jmcarp:
Builds on #494.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -29,3 +29,70 @@ jobs:
         run: make test
       - name: lint
         run: make lint
+  acceptance:
+    runs-on: ubuntu-latest
+    steps:
+      # Simulated omicron takes up a meaningful amount of disk space, and the
+      # hosted Github Actions runners don't offer much space. Clean up unused
+      # dependencies so that we don't run out of disk. Borrowed from
+      # https://carlosbecker.com/posts/github-actions-disk-space.
+      - name: "cleanup"
+        run: |
+          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc /opt/hostedtoolcache/CodeQL
+          sudo docker image prune --all --force
+          sudo docker builder prune -a
+      - uses: actions/checkout@v5
+      - uses: hashicorp/setup-terraform@v3
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+      - uses: docker/setup-compose-action@v1
+      - uses: astral-sh/setup-uv@v6
+      - name: install oxide cli
+        run: |
+          mkdir -p bin
+          wget https://github.com/oxidecomputer/oxide.rs/releases/download/v0.13.0+20250730.0.0/oxide-cli-x86_64-unknown-linux-gnu.tar.xz
+          tar xvf oxide-cli-x86_64-unknown-linux-gnu.tar.xz
+          mv oxide-cli-x86_64-unknown-linux-gnu/oxide bin
+          echo "$(pwd)/bin" >> $GITHUB_PATH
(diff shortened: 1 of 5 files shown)


acctest/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM rust
+
+SHELL ["/bin/bash", "-c"]
+
+RUN \
+  git clone https://github.com/oxidecomputer/omicron.git --depth 1 --branch jmcarp/prereqs-omit-sudo && \
```
jmcarp, line 6: Note: I had to make some small changes to the install scripts in omicron to get this to build. If we like this approach, I'll send a PR to omicron and stop pointing to my branch.

acctest/auth.py
```diff
@@ -0,0 +1,53 @@
+# /// script
```
jmcarp, line 1: This is a bit of a hack. To run these tests manually, the user would first run `oxide auth login`, but that's an interactive flow that requires a browser. To make this non-interactive, we add a script that simulates the device auth flow and prints a token. @david-crespo pointed out that we could instead support password login via the cli, with a warning not to use it in production, but IMO this is fine for the time being.

acctest/auth.py
```diff
@@ -0,0 +1,53 @@
...
+    token_resp = httpx.post(
+        f"{args.host}/device/token",
+        data={
+            "grant_type": "urn:ietf:params:oauth:grant-type:device_code",
+            "device_code": device_details["device_code"],
+            "client_id": str(u),
+        },
+    )
+    token_resp.raise_for_status()
+    return token_resp.json()["access_token"]
```
david-crespo, line 42: CLEAN

.github/workflows/build-test.yml
```diff
@@ -29,3 +29,54 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - uses: hashicorp/setup-terraform@v3
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+      - uses: docker/setup-compose-action@v1
+      - uses: astral-sh/setup-uv@v6
+      - name: install oxide cli
```
sudomateo, line 42: Just noting that we should probably consider creating a `setup-oxide` GitHub Action at some point.

.github/workflows/build-test.yml
```diff
@@ -29,3 +29,54 @@ jobs:
...
+          docker compose build
+          docker compose up --wait --wait-timeout 900
+      # We can't use `oxide auth login` here, since it requires a browser to
+      # complete the oauth device flow. Instead, fetch an auth token using a
+      # script that simulates the browser flow.
+      - id: auth-token
+        working-directory: acctest
+        run: |
+          uv run auth.py
+          echo "OXIDE_TOKEN=$(uv run auth.py)" >> $GITHUB_OUTPUT
```
sudomateo, line 64: What's the purpose of running `uv run auth.py` twice? Should we just capture the result as a variable and echo it instead?

acctest/Dockerfile
sudomateo: Oh, containers.

sudomateo (commented):
This is looking great! I know we have a few hacks in here for now but it'll be great to get acceptance tests running in automation while we sort those hacks out.

.github/workflows/build-test.yml
jmcarp: Oops, that was some leftover debug code. Dropping.

sudomateo (approved):
This is looking great. Just want to get #494 merged first before merging this.

sudomateo (changes requested):
Now that I merged #494 this needs conflict resolution and another review. Just a friendly reminder for @jmcarp.

PR #498, opened 2025-08-22
misc: use tfplugindocs to generate docs.

### LINKED ISSUE #490: Automate docs generation
### Overview

We have linters for automated docs quality checks, but AFAICT we haven't automated docs generation (unless I just missed it?). It looks like https://github.com/hashicorp/terraform-plugin-docs can do this for us, so I'll plan to test it out. Any concerns @sudomateo @karencfv?

### Implementation details

_No response_

### Anything else you would like to add?

_No response_

jmcarp:
Rather than constructing docs manually and running them through a linter, use tfplugindocs to generate them. We move examples into the examples/ directory, and override default templates in the templates/ directory. We also add new make targets to generate docs and check that the generated docs match the code.

Resolves #490.

Note: this is a low priority, but I wanted to fix the paper cut before investing more time in the provider.

Also note: the generated docs aren't identical to the current docs. I haven't gone through and tried to make them as similar as possible, but we can do more to minimize the diff if we want to.

### DIFF
--- a/internal/provider/utils.go
+++ b/internal/provider/utils.go
@@ -15,7 +15,7 @@ import (
 	"github.com/oxidecomputer/oxide.go/oxide"
 )
 
-// replaceBackticks replaces '' with `. It can be used to defined codeblocks in
+// replaceBackticks replaces ” with `. It can be used to defined codeblocks in
 // markdown raw strings.
 //
 //	var mdString = replaceBackticks(`this is a ''code'' block`)


sudomateo:
I really like this idea! I haven't used `tfplugindocs` so I want to better understand how it expects to use templates and spend some time updating our descriptions first so the generated documentation is consistent and useful. Agree it's low priority but I'll try to spend some time looking at this on the plane when I travel next week.

jmcarp:
Conflicts resolved. @lgfa29, thanks for adding the missing descriptions to the resource definitions. @sudomateo, this is ready for a look when you have time.

sudomateo:
> Conflicts resolved. @lgfa29, thanks for adding the missing descriptions to the resource definitions. @sudomateo, this is ready for a look when you have time.

Ack! Adding it to my list for the day. Have a few back-to-back meetings but later in the afternoon I'll review.

jmcarp:
Agreed that some of the assumptions the docs generator makes are silly, but also agreed that we can fix up the implications later on. Thanks for the review!

PR #474, opened 2025-07-21
resource(vpc_firewall_rules): support new api changes

sudomateo:
Pulled in the Go SDK changes from
https://github.com/oxidecomputer/oxide.go/pull/304 which contained changes to the VPC firewall rules APIs.

Updated the `oxide_vpc_firewall_rules` resource to account for these changes.

PR #501, opened 2025-08-27
resource(vpc_firewall_rules): add schema upgrade

### LINKED ISSUE #500: v0.13 provider unable to decode v0.12 state due to firewall_rules changes
### Preliminary checks

- [x] I am using the latest version, or the latest version that corresponds to my Oxide installation.
- [x] There is no open issue that reports the same problem.

### What was the expected behaviour

After upgrading their Oxide terraform provider to v0.13, a customer reported the following error:

```
Warning: Failed to decode resource from state

Error decoding
"oxide_vpc_firewall_rules.all_hosts[\"foo\"]" from prior
state: missing expected {

Warning: Failed to decode resource from state

Error: Unable to Read Previously Saved State for UpgradeResourceState

  with oxide_vpc_firewall_rules.all_hosts["foo"],
  on oxide_hosts.tf line 42, in resource "oxide_vpc_firewall_rules" "all_hosts":
  42: resource "oxide_vpc_firewall_rules" "all_hosts" {
```

This appears to be due to the updated `{ type = "<TYPE>" }` format now expected for firewall protocols.

### What is the current behaviour and what actions did you take to get there

Attempt to use a `state` file generated by provider v0.12.

### Provider version

v0.13

### Terraform version

1.12

### Operating system

Any

### Anything else you would like to add?

_No response_

lgfa29:
Implement the `ResourceWithStateUpgrade` interface on `vpcFirewallRulesResource` to support upgrading existing Terraform states to the new schema implemented in #474.

One important note is that this does not allow users to use new versions of the provider with previous versions of the Oxide API. It only updates existing Terraform states to avoid errors when users upgrade their clusters to a version that contains API changes.

Closes #500


internal/provider/resource_vpc_firewall_rules_v0.go
```diff
@@ -0,0 +1,724 @@
...
+	if diags.HasError() {
+		return nil, diags
+	}
+
+	upgradedProtocols := make([]vpcFirewallRuleProtocolFilterModelV01, len(protocols))
+	for i, p := range protocols {
+		upgradedProtocols[i] = vpcFirewallRuleProtocolFilterModelV01{
+			Type: types.StringValue(strings.ToLower(p.ValueString())),
+		}
+	}
```
lgfa29, line 395: Most of the code here is copy-paste and boilerplate to ensure Terraform has a stable schema when loading older state files. This is where the schema upgrade actually happens.

internal/provider/resource_vpc_firewall_rules.go
```diff
@@ -112,10 +113,24 @@ func (r *vpcFirewallRulesResource) ImportState(ctx context.Context, req resource
 	resource.ImportStatePassthroughID(ctx, path.Root("vpc_id"), req, resp)
 }
 
+// UpgradeState upgrades the Terraform state for the oxide_vpc_firewall_rules
+// resource from a previous schema version to the current version.
+//
+// Schema upgrades are not expected to be applied sequentially, since users are
+// allowed to jump to whatever new version they choose. When adding a new
+// version, you must ensure that each of the existing StateUpgrader functios
```
sudomateo, line 121:
```suggestion
// version, you must ensure that each of the existing StateUpgrader functions
```

internal/provider/resource_vpc_firewall_rules_test.go
```diff
@@ -611,3 +615,227 @@ func testAccFirewallRulesDestroy(s *terraform.State) error {
...
+          type  = "subnet"
+          value = "default"
+        }
+      ]
+    }
+  ]
+}
+`, ts.URL)
+
+	resource.UnitTest(t, resource.TestCase{
```
sudomateo, line 786: I see this block is `resource.UnitTest` and the following block is `resource.Test`. The code inside is mostly the same. Should both blocks be the same?

internal/provider/resource_vpc_firewall_rules_v0.go
```diff
@@ -0,0 +1,724 @@
...
+		}
+		resp.Diagnostics.Append(stateV01.Get(ctx, &modelV01)...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+	}
+
+	// Upgrade from schema v0.1 to latest.
+	newState := modelV01.upgrade()
+	resp.Diagnostics.Append(resp.State.Set(ctx, newState)...)
```
sudomateo, line 100: Just making sure I'm following this correctly. This v0.1 upgrade is really a no-op outside of type changes, correct? I see that v0.0 to v0.1 accounts for the new `type` field in the JSON but I do not see us doing anything special with v0.1 states. Just wanted to confirm.

sudomateo (commented):
Thank you for working on this! I have a few quick questions and comments before merging this.

internal/provider/resource_vpc_firewall_rules_test.go
lgfa29: Ah good catch. I was testing something else and found out about `UnitTest`, which I think works best here. We also don't need the `PreventPostDestroyRefresh` anymore (I was trying to test the scenarios where we get an error, but Terraform doesn't like when the state and schema don't match).

lgfa29:
Tested and validated the fix in a test environment.

### Environment upgrade test

1. Create test environment with R15.
2. Create an `oxide_vpc_firewall_rules` resource using provider v0.12.0.
3. Upgrade test environment to R16.
4. Upgrade provider to a binary running the code in this PR.
5. `terraform plan` and `terraform apply` run without errors and state schema for `oxide_vpc_firewall_rules` is upgraded to schema version 1.

### Provider upgrade test

1. Create test environment with R16.
2. Create an `oxide_vpc_firewall_rules` resource using provider v0.13.0.
3. Upgrade provider to a binary running the code in this PR.
4. `terraform plan` and `terraform apply` run without errors and state schema for `oxide_vpc_firewall_rules` is upgraded to schema version 1.

PR #505, opened 2025-09-02
added data_source_system_ip_pools

nickryand:
Added a new data source to lookup ip pools that at the System level. This is different from the current ip pools data source which looks up the pools within a silo.

### DIFF
--- a/.changelog/0.14.0.toml
+++ b/.changelog/0.14.0.toml
@@ -3,6 +3,8 @@ title = ""
 description = ""
 
 [[features]]
+title = "New data source"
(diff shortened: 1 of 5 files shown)


lgfa29 (approved):
Thank you for the PR @nickryand!

I pushed https://github.com/oxidecomputer/terraform-provider-oxide/pull/505/commits/d0b364ceb8f4158d34be547c5d958230afced67b to add some minor fixes and updates and also merged `main` to fix a conflict in the changelog file, but overall looks good.

Thanks again for the contribution.

PR #530, opened 2025-09-25
provider: use previous state for id attribute

sudomateo:
Updated resources to use the previous state for their `id` attribute if the resource has an `Update` method that can mutate the resource without changing the underlying Oxide ID.


internal/provider/resource_switch_port_settings.go
sudomateo, line 1: I need to test this resource since its `Update` method calls the `networking_switch_port_settings_create` API.

lgfa29 (approved):

PR #532, opened 2025-09-25
Support instance auto restart policy.

jmcarp:
Support new auto_restart_policy field on the instance resource, to be released in r17. Don't release until r17 is also released.


internal/provider/resource_instance.go
```diff
@@ -343,13 +355,14 @@ func (r *instanceResource) Create(ctx context.Context, req resource.CreateReques
...
 		Project: oxide.NameOrId(plan.ProjectID.ValueString()),
 		Body: &oxide.InstanceCreate{
-			Description: plan.Description.ValueString(),
-			Name:        oxide.Name(plan.Name.ValueString()),
-			Hostname:    oxide.Hostname(plan.HostName.ValueString()),
-			Memory:      oxide.ByteCount(plan.Memory.ValueInt64()),
-			Ncpus:       oxide.InstanceCpuCount(plan.NCPUs.ValueInt64()),
-			Start:       plan.StartOnCreate.ValueBoolPointer(),
-			UserData:    plan.UserData.ValueString(),
+			AutoRestartPolicy: oxide.InstanceAutoRestartPolicy(plan.AutoRestartPolicy.ValueString()),
```
sudomateo, line 358: There's no Go SDK change needed here? I thought this was made `nullable` upstream so we need to marshal this to `null` instead of omit it.

internal/provider/resource_instance.go
```diff
@@ -669,18 +687,26 @@ func (r *instanceResource) Update(ctx context.Context, req resource.UpdateReques
...
-				BootDisk: oxide.NameOrId(plan.BootDiskID.ValueString()),
-				Memory:   oxide.ByteCount(plan.Memory.ValueInt64()),
-				Ncpus:    oxide.InstanceCpuCount(plan.NCPUs.ValueInt64()),
+				Memory: oxide.ByteCount(plan.Memory.ValueInt64()),
+				Ncpus:  oxide.InstanceCpuCount(plan.NCPUs.ValueInt64()),
 			},
 		}
+		if !plan.AutoRestartPolicy.IsNull() {
+			autoRestartPolicy := oxide.InstanceAutoRestartPolicy(plan.AutoRestartPolicy.ValueString())
+			params.Body.AutoRestartPolicy = &autoRestartPolicy
```
sudomateo, line 704: Totally optional but if you wanted to avoid the type shuffling and variable creation. I know the type shuffling is a bit annoying.

```suggestion
			params.Body.AutoRestartPolicy = (*oxide.InstanceAutoRestartPolicy)(plan.AutoRestartPolicy.ValueStringPointer())
```

internal/provider/resource_instance.go
```diff
@@ -669,18 +687,26 @@ func (r *instanceResource) Update(ctx context.Context, req resource.UpdateReques
...
+		}
+		if !plan.BootDiskID.IsNull() {
+			bootDisk := oxide.NameOrId(plan.BootDiskID.ValueString())
+			params.Body.BootDisk = &bootDisk
```
sudomateo, line 708: Totally optional but if you wanted to avoid the type shuffling and variable creation. I know the type shuffling is a bit annoying.

```suggestion
			params.Body.BootDisk = (*oxide.NameOrId)(plan.BootDiskID.ValueStringPointer())
```

sudomateo (changes requested):
Thanks for working on this! The only thing that's missing is updating the tests to assert that this attribute is present.

I left some optional suggestions as well.

internal/provider/resource_instance.go
jmcarp: Updated the sdk and pulled in the latest version.

jmcarp:
Test updated!

sudomateo (approved):

Issue #515, opened 2025-09-11
vpc_firewall_rules: slow performance when specifying many rules


sudomateo:
The [vpc_firewall_rules_update](https://docs.oxide.computer/api/vpc_firewall_rules_update) API does not support per-rule updates. That means Terraform must read all rules, reconcile a diff, and PUT all rules back on every apply. The reconcile a diff phase can take minutes when managing hundreds of rules, which is unacceptable.

There was discussion on https://oxidecomputerhelp.zendesk.com/agent/tickets/340 where I noted that, outside of an API design change, we could consider changing the type of `rules` from `SetNestedAttribute` to `ListNestedAttribute`. Below is the testing that confirms this consideration may be a valid path.

I created a configuration that manages 300 VPC firewall rules.

```hcl
data "oxide_project" "main" {
  name = "matthewsanabria"
}

resource "oxide_vpc" "example" {
  project_id  = data.oxide_project.main.id
  description = "example"
  name        = "example"
  dns_name    = "example"
}

resource "oxide_vpc_firewall_rules" "example" {
  vpc_id = oxide_vpc.example.id
  rules = [for port in range(8000, 8300):
    {
      action      = "allow"
      description = format("Rule for %d.", port)
      name        = format("rule-%d", port)
      direction   = "inbound"
      priority    = 50
      status      = "enabled"
      filters = {
        ports     = [port]
        protocols = [{
          type = "tcp",
        }]
      },
      targets = [
        {
          type  = "subnet"
          value = "default"
        }
      ]
    }
  ]
}
```

After the initial apply I ran a subsequent no-op apply and it took 146 seconds which is unacceptable.

```
> time terraform apply
data.oxide_project.main: Reading...
data.oxide_project.main: Read complete after 0s [id=5476ccc9-464d-4dc4-bfc0-5154de1c986f]
oxide_vpc.example: Refreshing state... [id=92b5dae2-9bb3-4990-b817-ac1007c1d8fc]
oxide_vpc_firewall_rules.example: Refreshing state... [id=80ec78f8-ff21-458a-b596-4cd2f38d53a5]

No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

________________________________________________________
Executed in   70.82 secs    fish           external
   usr time  146.25 secs    0.85 millis  146.25 secs
[...]

PR #543, opened 2025-10-16
vpc_firewall_rules: switch schema to increase performance

sudomateo:
Switched the schema for the `rules` attribute from `schema.SetNestedAttribute` to `schema.MapNestedAttribute` to increase performance when working with large numbers of firewall rules.

Closes https://github.com/oxidecomputer/terraform-provider-oxide/issues/515.

### DIFF
--- a/.changelog/0.16.0.toml
+++ b/.changelog/0.16.0.toml
@@ -1,6 +1,6 @@
 [[breaking]]
-title = ""
-description = ""
+title = "`oxide_vpc_firewall_rules`"
+description = "Changed the schema of the `rules` attribute from a set to a hashmap for better performance. Refer to the resource documentation for how to upgrade your configuration. [#543](https://github.com/oxidecomputer/terraform-provider-oxide/pull/543)"
 
 [[features]]
 title = ""


internal/provider/resource_vpc_firewall_rules.go
```diff
@@ -388,19 +383,14 @@ func (r *vpcFirewallRulesResource) Create(ctx context.Context, req resource.Crea
...
 	plan.TimeCreated = types.StringNull()
 	plan.TimeModified = types.StringNull()
 	if len(plan.Rules) > 0 {
-		plan.TimeCreated = plan.Rules[0].TimeCreated
-		plan.TimeModified = plan.Rules[0].TimeModified
+		for _, rule := range plan.Rules {
+			plan.TimeCreated = rule.TimeCreated
+			plan.TimeModified = rule.TimeModified
+			break
+		}
```
sudomateo: We only need `time_created` and `time_modified` values from any rule. Perhaps we can extract this into a helper function since we're doing this in a few places now.

internal/provider/resource_vpc_firewall_rules_v01.go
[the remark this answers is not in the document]
lgfa29: Fixed in https://github.com/oxidecomputer/terraform-provider-oxide/pull/543/commits/e81f575c9b674e2433d9b5b5475771a2c48333c3.

docs/resources/vpc_firewall_rules.md
```diff
@@ -6,9 +6,71 @@ description: |-
   This resource manages VPC firewall rules.
   !> Firewall rules defined by this resource are considered exhaustive and will
   overwrite any other firewall rules for the VPC once applied.
-  !> Setting the rules attribute to [] will delete all firewall rules for the
+  !> Setting the rules attribute to {} will delete all firewall rules for the
   VPC which may cause undesired network traffic. Please double check the firewall
   rules when updating this resource.
+  Migrating rules
```
sudomateo: Don't have to do this now but we should probably stop using the large description for this YAML front matter.

docs/resources/vpc_firewall_rules.md
sudomateo: Let's note that we have to remove the `name` attribute when migrating to the new schema. I imagine people upgrading will forget to do that and then try and apply and get this error.

```
╷
│ Error: Invalid Configuration for Read-Only Attribute
│
│   with oxide_vpc_firewall_rules.example,
│   on main.tf line 12, in resource "oxide_vpc_firewall_rules" "example":
│   12: resource "oxide_vpc_firewall_rules" "example" {
│
│ Cannot set value for this attribute as the provider has marked it as read-only. Remove the configuration line setting the value.
│
│ Refer to the provider documentation or contact the provider developers for additional information about configurable and read-only attributes that
│ are supported.
╵
```

lgfa29: I made the migration steps more explicit and removed most of the unnecessary content from the migration example to make the fact that `name` was removed more explicit. I tried to use the `diff` syntax highlight, but it doesn't seem to be supported by the Terraform docs, and it can also be hard to read if you're not used to the `-/+` syntax. 

Here's the rendered preview of the new content:
  
Screen Shot 2025-10-27 at 18 21 19

docs/resources/vpc_firewall_rules.md
```diff
@@ -6,9 +6,49 @@ description: |-
...
+  Previous versions of this resource stored firewall rules in a set. This
+  resulted in slow plans in environments with a significant number of rules.
+  Newer versions store the rules in a map for better performance, but this change
```
jmcarp: Your screenshot looks good, but did we lose some formatting here? Should this be a numbered list?

internal/provider/function_to_vpc_firewall_rules_map.go
```diff
@@ -0,0 +1,264 @@
...
+			// so we "hack" around it by receiving a JSON encoded string of the
+			// HCL configuration.
+			function.StringParameter{
+				Name: "rules_json",
+				MarkdownDescription: replaceBackticks(`
+JSON encoded string of the ''rules'' set. Use the
+[''jsonencode''](https://developer.hashicorp.com/terraform/language/functions/jsonencode)
+function to convert the existing ''rules'' set to a JSON string.
+`),
+			},
```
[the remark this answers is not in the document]
jmcarp: We'll deprecate this function in a future version, right? Especially if that's the case, I agree that it's not worth spending time to figure out another approach.

docs/resources/vpc_firewall_rules.md
sudomateo: Agreed. I'd like to take a look at this in a future change.

PR #549, opened 2025-11-10
fix: simplify omicron configuration

sudomateo:
Simplified the Omicron configuration for integration tests by using the configuration from the upstream repository and layering in the configuration we need. I also updated the container files to be compatible with Podman.

Fixes SSE-52.


jmcarp (approved):
Nice, this is much better.

PR #575, opened 2025-12-17
ci: fix backport permissions

lgfa29:
GitHub has an undocumented permission called `workflows` that is required to allow modifying workflow files.

`actions/create-github-app-token` is documented to import the GitHub App permissions by default, but from my tests it doesn't seem to import `workflows`, so permissions need to be set explicitly.


lgfa29:
This is a trivial change that I tested in the backport test repo, so merging it before I forget about it again 😅

PR #579, opened 2025-12-20
feat(oxide_instance): read-only attribute to show instance external ips

sudomateo:
Added a new computed attribute to `oxide_instance` that exposes the external IP addresses for the instance. This should allow users to use a provisioner with the `oxide_instance` resource, though further testing is needed.

Amp-Thread: https://ampcode.com/threads/T-019b39f4-0dc1-708b-9b75-ba69bf65c73c

Closes SSE-75.
Relates to SSE-64.

-----

### Pull request checklist

### DIFF
--- a/docs/resources/instance.md
+++ b/docs/resources/instance.md
@@ -130,6 +130,7 @@ Maximum 32 KiB unencoded data.
 
 ### Read-Only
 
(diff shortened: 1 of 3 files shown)


sudomateo:
I'll come back to this in a future pull request after we settle on #597 and release 18 things.

PR #592, opened 2026-01-13
ci: install oxide CLI using the specified branch

lgfa29:
The `acc-test-setup.sh` script uses the `oxide` CLI to seed the simulated environment with the resources required to run the acceptance tests, so it needs to be in-sync with the Omicron version being used in the simulator.

Raising as a separate PR so it's easier to backport.


jmcarp (approved):
LGTM.

PR #593, opened 2026-01-13
deps: update Go SDK with latest changes for R18

lgfa29:
This PR is the first of a series intended to add support for release R18 features.

This initial work just updates the SDK version and adapts the internal logic to the new API structure. It does not add new feature and does not modify any resource schema.

Most importantly, this PR does not add support for IPv6. Network cards with IPv6 are ignored and it's not possible to create IPv6 network interfaces or external IPs.

Relates to SSE-141

### DIFF
--- a/internal/provider/resource_instance.go
+++ b/internal/provider/resource_instance.go
@@ -6,6 +6,7 @@ package provider
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"time"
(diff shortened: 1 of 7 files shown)


internal/provider/resource_instance.go
```diff
@@ -1316,7 +1349,15 @@ func newExternalIPsOnCreate(externalIPs []instanceResourceExternalIPModel) []oxi
 
 		if ip.Type.ValueString() == string(oxide.ExternalIpCreateTypeEphemeral) {
 			if ip.ID.ValueString() != "" {
-				eIP.Pool = oxide.NameOrId(ip.ID.ValueString())
+				eIP.PoolSelector = oxide.PoolSelector{
+					Type: oxide.PoolSelectorTypeExplicit,
+					Pool: oxide.NameOrId(ip.ID.ValueString()),
```
jmcarp: I can go look it up, but is the pool here the string value of the ip address? That seems like an odd interface.

PR #597, opened 2026-01-16
instance: update external IPs schema

lgfa29:
Add IPv6 support for external IPs in the `oxide_instance` resource. This is a breaking change that modifies the resource schema to model the change in spec between ephemeral and floating IPs. The migration code handles updates so that applies are no-ops.

Relates to SSE-141

-----

### Pull request checklist

### DIFF
--- a/internal/provider/resource_instance.go
+++ b/internal/provider/resource_instance.go
@@ -140,13 +140,10 @@ func (ip *instanceResourceExternalIPModel) Empty() bool {
 type instanceResourceEphemeralIPModel struct {
 	PoolID    types.String `tfsdk:"pool_id"`
 	IPVersion types.String `tfsdk:"ip_version"`
-	IP        types.String `tfsdk:"ip"`
 }
 
 type instanceResourceFloatingIPModel struct {
-	ID   types.String `tfsdk:"id"`
-	Name types.String `tfsdk:"name"`
-	IP   types.String `tfsdk:"ip"`
+	ID types.String `tfsdk:"id"`
 }
 
 var instanceResourceAttachedNICType = types.ObjectType{}.WithAttributeTypes(
@@ -561,10 +558,6 @@ This resource manages instances.
 										),
 									},
 								},
-								"ip": schema.StringAttribute{
-									Computed:    true,
-									Description: "The external ephemeral IP attached to the instance.",
-								},
 							},
 						},
 					},
@@ -576,14 +569,6 @@ This resource manages instances.
 								"id": schema.StringAttribute{
 									Required:    true,
 									Description: "The external floating IP ID.",
-								},
-								"ip": schema.StringAttribute{
-									Computed:    true,
-									Description: "The external floating IP attached to the instance.",
-								},
-								"name": schema.StringAttribute{
-									Computed:    true,
-									Description: "The name of the external floating IP attached to the instance.",
 								},
 							},
 						},
@@ -889,12 +874,6 @@ func (r *instanceResource) Create(
 	for i, ip := range instExternalIPs.Ephemeral {
 		plan.ExternalIPs.Ephemeral[i].PoolID = ip.PoolID
 		plan.ExternalIPs.Ephemeral[i].IPVersion = ip.IPVersion
-		plan.ExternalIPs.Ephemeral[i].IP = ip.IP
-	}
-
-	for i, ip := range instExternalIPs.Floating {
-		plan.ExternalIPs.Floating[i].IP = ip.IP
-		plan.ExternalIPs.Floating[i].Name = ip.Name
 	}
 
 	// Populate Computed attribute values about network interfaces.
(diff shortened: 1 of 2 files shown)


lgfa29:
@sudomateo I believe this also closes https://github.com/oxidecomputer/terraform-provider-oxide/pull/579?

internal/provider/resource_instance.go
```diff
@@ -502,36 +521,75 @@ This resource manages instances.
 					},
 				},
 			},
-			"external_ips": schema.SetNestedAttribute{
+			"external_ips": schema.SingleNestedAttribute{
```
sudomateo: I like the flattened schema overall. In the schema, `ephemeral.pool_id` maps to `explicit` and `ephemeral.ip_version` maps to `auto`. Do you think this is a clear enough representation for the operator to understand one is explicit and the other auto? Should we take the time now to model the API structure directly?

sudomateo (commented):
This looks good to me. I like using a map to model enum variants. It's the cleanest we can do right now without creating a custom type. I left a question about whether we want to model the API design fully that we can briefly discuss before moving forward.

lgfa29:
> > @sudomateo I believe this also closes #579?
> 
> It likely can if these changes unblock the ability to use a `provisioner`. The goals of #579 are twofold.
> 
>     1. Unblock `provisioner`.
> 
>     2. Expose all external IP information (e.g., SNAT IPs).

Ah, I missed the provisioner part. But I think it should work fine, I will give it a try tomorrow. We should also be able to add SNAT now, it would just be a new list in the `external_ips` object.

internal/provider/resource_instance.go
lgfa29: After spending a couple of hours this morning trying to split `external_ips` into a read (`attached_external_ips`) and a write attribute, I kept hitting into inconsistent state errors, similar to the bug I described for updating network interfaces IP stack.

I think the crux of the issue is that this kind of split makes updates and state reconciliation harder to deal with because we now need to correlate the two attributes.

For example, image you start with something like this:

```hcl
external_ips = {
  ephemeral = [
    { ip_version = "v4" },
  ]
}
```

Ideally, the split attribute state would look like this:

```json
{
  "external_ips": {
    "ephemeral": [
      "ip_version": "v4"
    ]
  },
  "attached_external_ips": {
    "ephemeral": {
      "v4": {
        "ip": "...",
        "ip_pool_id": "...",
      }
    }
  }
}
```

But now if the instance gets updated out-of-band to, for example, use IPv6 for the ephemeral IP, the only state value that is going to change on `Read()` is the `attached_external_ips`, which will then be inconsistent with `external_ips`.

So the challenge is then to reconstruct what `ephemeral_ips` should look like as if the user wrote the config in the first place. But this requires some care. In this example, `Read()` would only have access to `ip` and `ip_pool_id` from the IP, and it needs to figure out that the value for `ephemeral_ips` should have `ip_version` and not `pool_id`, otherwise the user will be required to set `pool_id` to avoid infinite diffs.

All of this to say that I think I will hold off adding the computed attributes to `ephemeral_ips` for now. So it will support creating IPv6 addresses, but you will still need the data source to read the actual IP values, keeping the existing behaviour in-place.

I will then look into the bug we have with network interfaces and see if I can help us with external IPs as well.

internal/provider/resource_instance_test.go
```diff
@@ -1831,8 +2006,41 @@ func checkResourceInstanceIPUpdate1(resourceName, instanceName string) resource.
...
+		),
+		resource.TestCheckTypeSetElemNestedAttrs(
+			resourceName,
+			"external_ips.ephemeral.*",
+			map[string]string{"ip_version": "v6"},
+		),
+	}...)
+}
+
+func checkResourceInstanceIPUpdate1SingleStack(
```
jmcarp: This is a general question, not really specific to this test, but: to make the intention clearer, I wonder if we could just assert on the attributes that we're interested in, which here I think would be `external_ips/ephemeral/...`. Is it useful to test that e.g. the memory and cpu didn't change when we updated the external ips attribute? If yes, maybe we can visually group the assertions into things that shouldn't change (boilerplate, probably copied from a previous test) and things that should. Just a nit.

jmcarp (approved):
Small question about a test we might want to add, but nothing blocking. LGTM.

internal/provider/resource_instance.go
```diff
@@ -570,6 +613,108 @@ Maximum 32 KiB unencoded data.`,
 	}
 }
 
+func (r *instanceResource) UpgradeState(ctx context.Context) map[int64]resource.StateUpgrader {
```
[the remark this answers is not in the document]
lgfa29: Ideally yes, in practice testing the `oxide_instance` resource upgrade is quite more complicated than say the [`oxide_vpc_firewall_rules`](https://github.com/oxidecomputer/terraform-provider-oxide/blob/a7fc28f0a28ef961a7c9536eba14d6e1ea36452f/internal/provider/resource_vpc_firewall_rules_test.go#L874) for example because it makes several API calls.

Long-term, I think we may be able to automate upgrade paths using something like [`testcontainers-go`](https://golang.testcontainers.org/quickstart/) but I think that will require quite a bit of work to get right.

PR #598, opened 2026-01-16
Bump oxide.go and handle interface types.

jmcarp:
Update to use https://github.com/oxidecomputer/oxide.go/pull/359 to work through how the changes will affect the provider. Marking as a draft while we finish the upstream PR. Overall, some of the code got more verbose, because we have to `switch` over all known variant types, but we also dropped the json marshal/unmarshal hack, and type safety is improved.

-----

### Pull request checklist


internal/provider/resource_instance.go
```diff
@@ -530,16 +529,8 @@ func (r *instanceResource) Create(ctx context.Context, req resource.CreateReques
...
-			stack, err := ipStackAsIPv4Stack(nic.IpStack)
-			if err != nil {
-				resp.Diagnostics.AddError(
-					"Unable to read instance network interface:",
-					"Error: "+err.Error(),
-				)
-				continue
-			}
-			ip = stack.Ip
+		if v4, ok := nic.IpStack.Value.(*oxide.PrivateIpStackV4); ok {
```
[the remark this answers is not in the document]
sudomateo: I like this much better but am slightly struggling to understand why we can use the type assertion here at the top level rather than needing to create helper functions like shown in https://github.com/oxidecomputer/terraform-provider-oxide/pull/598/changes#r2700298578. This might just be late Friday brain where I'm missing something small.

jmcarp:
I'm pretty happy with these changes now. We don't actually have a huge number of oneOf types in the api, so the changes only touch a few files in the terraform provider. The changes fall into a few categories:

* Trivial changes, like s/Type/Type()/.
* New constructors for oneOf types when preparing the body of an api request.
* Error handling for those constructors, since they can return errors if we encounter an unexpected variant type.

internal/provider/resource_vpc_firewall_rules.go
```diff
@@ -706,13 +735,12 @@ func newVPCFirewallRulesModel(rules []oxide.VpcFirewallRule) (map[string]vpcFire
 			// We can safely dereference rule.Priority as it's a required field
 			Priority:     types.Int64Value(int64(*rule.Priority)),
 			Status:       types.StringValue(string(rule.Status)),
-			Targets:      newTargetsModelFromResponse(rule.Targets),
+			Targets:      targets,
```
lgfa29: I'm necessarily opposed to the change, but it seems unnecessary in this case since the response from `newTargetsModelFromResponse()` is not being processed?

internal/provider/resource_vpc_router_route.go
```diff
@@ -419,3 +447,59 @@ func (r *vpcRouterRouteResource) Delete(ctx context.Context, req resource.Delete
 
 	tflog.Trace(ctx, fmt.Sprintf("deleted VPC router route with ID: %v", state.ID.ValueString()), map[string]any{"success": true})
 }
+
+// newRouteDestination creates a RouteDestination from type and value strings.
+func newRouteDestination(typeStr, value string) (oxide.RouteDestination, error) {
+	switch typeStr {
+	case "ip":
```
lgfa29: Same comment as above wrt using SDK values instead of strings.

lgfa29 (approved):
Changes look like a nice improvement. I will rebase my WIP branches once this is merged to get a hands-on experience with the new SDK.

I also wondered about the generators, but couldn't think of a good way to handle them. I wouldn't say they're Terraform specific, but this case is somewhat particular because all variants have a similar structure, so it's easy to create a `New*(string, string)` function. If any of the variants were a little different, the constructor would start to become quite messy (like having arguments that are only used with specific types).

PR #604, opened 2026-01-27
Refactor and cache docker image for faster builds.

jmcarp:
* Use a builder image to build the relevant binaries, then copy to the output image. This will avoid compiling rust binaries on every run.
* Check for cached images based on VERSION_OMICRON if available, other than the cron task, which always uses omicron main.

### DIFF
--- a/acctest/Dockerfile
+++ b/acctest/Dockerfile
@@ -1,12 +1,45 @@
-FROM rust
+# Builder image
+FROM rust:bookworm AS builder
 
-ARG OMICRON_BRANCH="main"
+ARG OMICRON_SHA
 
-RUN /bin/bash -c "git clone https://github.com/oxidecomputer/omicron.git --branch '${OMICRON_BRANCH}' --depth 1 && \
-  cd omicron && \
-  source env.sh && \
-  ./tools/install_builder_prerequisites.sh -y -s"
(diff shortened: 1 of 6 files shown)


acctest/Dockerfile
```diff
@@ -1,12 +1,48 @@
...
+FROM debian:bookworm-slim
+
+RUN apt-get update && apt-get install -y \
+    ca-certificates \
+    libpq5 \
+    libsqlite3-0 \
+    libssl3 \
+    libxmlsec1-openssl \
+    curl \
+    && rm -rf /var/lib/apt/lists/*
```
sudomateo: Should we just run the `tools/install_runner_prerequisites.sh` script to get these packages?

jmcarp:
> I left a small question but I did want to also ask whether we should put some workflow dependencies in place so that `acceptance-sim.yml` depends on `build-test.yml`.

IIUC this is handled by the `needs` block in the [acceptance stanza](https://github.com/oxidecomputer/terraform-provider-oxide/pull/604/changes#diff-063ca77e012f959eba648db60e6868875fd1e70e5f5ac081193d848f8d61ef32R108):

```
  acceptance:
    needs: [omicron-version, publish-image]
    uses: "./.github/workflows/acceptance-sim.yml"
  ...
```

PR #609, opened 2026-02-02
chore: print omicron logs on acctest failure.

jmcarp:
-----

### Pull request checklist

### DIFF
--- a/.github/workflows/acceptance-sim.yml
+++ b/.github/workflows/acceptance-sim.yml
@@ -72,11 +72,7 @@ jobs:
         working-directory: acctest
         env:
           TEST_ACC_DOCKER_TAG: ${{ inputs.omicron-sha }}
-        run: |
-          if ! docker compose up --wait --wait-timeout 1500; then
-            docker compose logs
-            exit 1
-          fi
+        run: docker compose up --wait --wait-timeout 1500
       # We can't use `oxide auth login` here, since it requires a browser to
       # complete the oauth device flow. Instead, fetch an auth token using a
       # script that simulates the browser flow.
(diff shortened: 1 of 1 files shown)


.github/workflows/acceptance-sim.yml
```diff
@@ -117,3 +117,7 @@ jobs:
           OXIDE_TOKEN: ${{ steps.auth-token.outputs.OXIDE_TOKEN }}
           OXIDE_TEST_IP_POOL_NAME: default
           OXIDE_SILO_DNS_NAME: "*.sys.oxide-dev.test"
+      - name: print omicron logs
+        if: failure()
+        working-directory: acctest
+        run: docker compose logs
```
lgfa29: Does STDOUT and STDERR actually contain useful information that helped you debug something? 

I think uploading the Omicron files from `/tmp/omicron-dev-omicron-dev.*.log` may be more helpful.

jmcarp: Ok, I'm revising this a bit to print both container logs and log files, and upload to workflow artifacts, unconditionally.

.github/workflows/acceptance-sim.yml
```diff
@@ -117,3 +111,25 @@ jobs:
...
+      - name: extract omicron log files
+        if: always()
+        continue-on-error: true
+        working-directory: acctest
+        run: |
+          docker compose exec -T omicron-dev sh -c 'tar -cf - /tmp/omicron-dev*.log' | tar -xf -
+          for f in tmp/*.log; do
+            echo "=== Last 100 lines of $f ==="
+            tail -100 "$f"
+          done
```
lgfa29: I think it may be better to mount  the container's `/tmp` to the host and upload the full log. The omicron logs are _very_ chatty, so 100 lines may not be enough to debug problems.

PR #614, opened 2026-02-03
Add subnet pool resource and data source.

jmcarp:
Modeled after ip pool resource and data source. Note that we also bump oxide.go to capture recent changes in the api.

Relates to SSE-141.

-----

### Pull request checklist


.changelog/0.18.0.toml
```diff
@@ -22,6 +22,14 @@ description = "Added support for the `config_dir` and `insecure_skip_verify` pro
...
 description = "Added support for the `disk_type` attribute for the `oxide_disk` resource. [#587](https://github.com/oxidecomputer/terraform-provider-oxide/pull/587)."
 
+[[features]]
+
+title = "New resource"
+description = "`oxide_subnet_pool` [#614](https://github.com/oxidecomputer/terraform-provider-oxide/pull/614)."
+
+title = "New data source"
+description = "`oxide_subnet_pool` [#614](https://github.com/oxidecomputer/terraform-provider-oxide/pull/614)."
+
```
lgfa29:
```suggestion
[[features]]
title = "New resource"
description = "`oxide_subnet_pool` [#614](https://github.com/oxidecomputer/terraform-provider-oxide/pull/614)."

[[features]]
title = "New data source"
description = "`oxide_subnet_pool` [#614](https://github.com/oxidecomputer/terraform-provider-oxide/pull/614)."
```

I think you need separate item headers.

internal/provider/resource_subnet_pool.go
```diff
@@ -0,0 +1,640 @@
...
+				Validators: []validator.String{
+					stringvalidator.OneOf("v4", "v6"),
+				},
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"members": schema.ListNestedAttribute{
+				Optional:    true,
+				Computed:    true,
```
lgfa29: Lists can cause infinite applies if the order they're returned from the API don't match what's in the config. For example, if you switch the mermber order in `testResourceSubnetPoolUpdateConfig ` you get this error:

```
--- FAIL: TestAccResourceSubnetPool_full (6.28s)
    resource_subnet_pool_test.go:21: Step 2/7 error: Error running apply: exit status 1

        Error: Provider produced inconsistent result after apply

        When applying changes to oxide_subnet_pool.test, provider
        "provider[\"registry.terraform.io/hashicorp/oxide\"]" produced an unexpected
        new value: .members[0].max_prefix_length: was cty.NumberIntVal(30), but now
        cty.NumberIntVal(28).

        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.

        Error: Provider produced inconsistent result after apply

        When applying changes to oxide_subnet_pool.test, provider
        "provider[\"registry.terraform.io/hashicorp/oxide\"]" produced an unexpected
        new value: .members[0].subnet: was cty.StringVal("198.51.100.0/24"), but now
        cty.StringVal("192.0.2.0/24").

        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.

        Error: Provider produced inconsistent result after apply

        When applying changes to oxide_subnet_pool.test, provider
        "provider[\"registry.terraform.io/hashicorp/oxide\"]" produced an unexpected
        new value: .members[1].max_prefix_length: was cty.NumberIntVal(28), but now
        cty.NumberIntVal(30).

        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.

        Error: Provider produced inconsistent result after apply

        When applying changes to oxide_subnet_pool.test, provider
        "provider[\"registry.terraform.io/hashicorp/oxide\"]" produced an unexpected
        new value: .members[1].min_prefix_length: was cty.NumberIntVal(26), but now
        cty.NumberIntVal(24).

        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.

        Error: Provider produced inconsistent result after apply

        When applying changes to oxide_subnet_pool.test, provider
        "provider[\"registry.terraform.io/hashicorp/oxide\"]" produced an unexpected
        new value: .members[1].subnet: was cty.StringVal("192.0.2.0/24"), but now
        cty.StringVal("198.51.100.0/24").

        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.
```

So this may be better model as a set, but that comes with its own little fun surprises, specially when the set elements have computed values. Terraform indexes the elements by hash, so if their values change after an apply you get an error.

There are a few options here:
1. Try using a set. All element attributes are simple values, which Terraform seem to handle better than complex types. This is not quite future-proof, as we may run into problems if a complex attribute is added in the future, but maybe that's OK for now.
2. Use a plan modifier to sort the list based on the user config file. I'm not 100% sure this will work, but that's something I thought about when dealing with the network interface updates.
3. Split the write and read parts into separate attributes. This is what I did in https://github.com/oxidecomputer/terraform-provider-oxide/pull/591, but may be an overkill here.

PR #622, opened 2026-02-09
Validate subnet cidrs.

jmcarp:
* Add a subnet cidr validator, which optionally accepts an address type.
* Move existing ip config validator to utils and backfill tests.

Note: this isn't an attempt to validate all ip address and cidr fields in the provider. If we find these validators useful, we can expand validation coverage in a separate patch.

-----

### Pull request checklist


internal/provider/utils.go
lgfa29: Same here and for its tests.

PR #627, opened 2026-02-10
chore: skip fail-fast on terraform/tofu test matrix.

jmcarp:
Sometimes terraform and opentofu tests fail for different reasons (usually test flakes). It's useful to see whether only one branch is failing (probably a flake) or both (more likely a true regression). This patch sets `fail-fast: false` so that we allow both branches to run to completion even if one fails.

-----

### Pull request checklist


lgfa29 (approved):
I will be forever annoyed that `fail-fast: true` is the default.

PR #632, opened 2026-02-11
Fix and refactor vpc subnet test.

jmcarp:
The TestAccCloudResourceVPCSubnet_full test has been failing probabilistially: on recreating the subnet, we also change the v4 cidr, and this fails when terraform happens to attempt to create the new subnet before the old one is destroyed. There's no particular value in testing the new v4 cidr, so we use the same cidr throughout as of this patch.

We also refactor the subnet test suite for clarity, using a single template and changing its parameters to show intent. We also drop unnecessary template variables and random resource names, which we don't need.

@lgfa29 suggested getting rid of these unnecessary template variables, and I agree. Sometimes I have trouble figuring out the intent of an existing acceptance test, and I think the vpc subnet tests read better after this change. Once we agree on a consistent testing style, we may want to try having claude rewrite the existing tests.

### DIFF
--- a/internal/provider/resource_vpc_subnet_test.go
+++ b/internal/provider/resource_vpc_subnet_test.go
@@ -15,32 +15,34 @@ import (
 	"github.com/oxidecomputer/oxide.go/oxide"
 )
 
-type resourceVPCSubnetConfig struct {
-	BlockName        string
-	SubnetName       string
-	SupportBlockName string
-	VPCBlockName     string
-	VPCName          string
+type vpcSubnetTestConfig struct {
+	SubnetName  string
+	Description string
+	IPv4Block   string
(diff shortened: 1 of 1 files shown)


jmcarp:
> I think it has always been one subnet that is updated on each test case.

I think there are two distinct subnets that use similar templates with different values for blockName. Compare config (using resourceVPCSubnetConfigTpl) and config2 (using resourceVPCSubnetIPv6ConfigTpl). When we run the fourth step in Steps (index 3), terraform wants to destroy the first subnet (named blockName) and create the second (blockName2), which is where we get the non-deterministic confict.

Am I reading this wrong? Like I said at the top, I found the existing test kind of hard to read.

PR #634, opened 2026-02-11
instance: fix network interface plan modifier

lgfa29:
The instance `mac_address` is also a computed attribute, so it needs to be set to `unknown` if the network interface is recreated.

I tried to create a reflect-based approach to handle this automatically, but the code was a bit too much for what it is doing. I don't expect we would change the schema too often, so I found a more explicit approach better. We should reconsider this decision If we find ourselves running into problems keeping this function updated though.


jmcarp (approved):
This LGTM. Would it be worthwhile (or even possible) to write a deterministic regression test for this?

PR #652, opened 2026-02-20
instance: fix state migration from v0

lgfa29:
Previous implementation of the `oxide_instance` resource state migration from v0 used model structs from an unreleased commit instead of the v0.17.0 tag, so it contained a lot of unnecessary fields.

Terraform ignores them when parsing real state into these structs, and we currently ignore them for the migration logic, but they could cause problems if we do try to access them in the future.

-----

### Pull request checklist

No changelog since this is an internal fix.

### DIFF
--- a/internal/provider/resource_instance.go
+++ b/internal/provider/resource_instance.go
@@ -700,24 +700,16 @@ func (r *instanceResource) UpgradeState(ctx context.Context) map[int64]resource.
 						TimeCreated:  oldNIC.TimeCreated,
 						TimeModified: oldNIC.TimeModified,
 						VPCID:        oldNIC.VPCID,
+
+						// New attribute added in schema v1.
+						//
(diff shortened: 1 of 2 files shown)


internal/provider/resource_instance.go
```diff
@@ -700,24 +700,16 @@ func (r *instanceResource) UpgradeState(ctx context.Context) map[int64]resource.
 						TimeCreated:  oldNIC.TimeCreated,
 						TimeModified: oldNIC.TimeModified,
 						VPCID:        oldNIC.VPCID,
-					}
-
-					if oldNIC.IPConfig != nil {
```
lgfa29: We don't have any user impact with v0.18.0 having the wrong logic because this will never be true (`oldNIC.IPConfig` will always be `nil` when parsing a v0 schema).

sudomateo (approved):

PR #658, opened 2026-02-25
Fix regression on icmp protocol config.

jmcarp:
As of main, configuring an icmp firewall rule with no icmp type or code set results in an error from the api:

```hcl
resource "oxide_vpc_firewall_rules" "example" {
  rules = {
    "allow-icmp" = {
      filters = {
        protocols = [{ type = "icmp" }]
      }
      ...
    }
  }
}
```

```
unable to parse JSON body: rules[0].filters.protocols[0].value.icmp_type:
invalid type: null, expected u8 at line 1 column 166
```

This patch fixes the bug by omitting the protocol value field if the icmp type and code are both empty. Note that there's arguably also a bug in the oxide sdk, such that icmp type is a required field but modeled as nullable (`*int`); we'll consider a separate fix there as well.

We also add a regression test, and refactor existing firewall rules tests to avoid unnecessary templating of terraform resource names.

-----

### Pull request checklist

### DIFF
--- a/internal/provider/function_to_vpc_firewall_rules_map_test.go
+++ b/internal/provider/function_to_vpc_firewall_rules_map_test.go
@@ -14,17 +14,11 @@ import (
 )
 
 func TestAccFunctionToVPCFirewallRulesMap_full(t *testing.T) {
-	blockName := newBlockName("firewall_rules")
-	supportBlockName := newBlockName("support")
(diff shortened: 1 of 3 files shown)


sudomateo (approved):
Looks great! Thanks for catching this.

We have R19 on the horizon so I'm okay either backporting this as a patch or waiting for the next release. Any preference?

PR #703, opened 2026-04-07
Bump oxide.go to latest.

jmcarp:
Bump the sdk to latest to capture changes in nexus. This pulls in a breaking change: the external subnet resource's prefix_len is renamed to prefix_length. We add a quick state migration to stay in sync with the upstream api.

-----

### Pull request checklist


sudomateo (approved):
This looks good to me!

lgfa29 (changes requested):
Sorry you went through a state migration path, but I think we need to handle this change differently.

PR #732, opened 2026-04-08
misc: repository refactor phase 1

sudomateo:
Refactored the repository according to phase 1 from SSE-266.

* Created the `internal/provider/shared` package and moved `utils.go` and `planmodifiers.go` within, exporting the necessary identifiers for use by other packages.

* Exported identifiers in `testutils.go` for use by other packages.

Part of SSE-266.

### DIFF
--- a/internal/provider/shared/utils.go
+++ b/internal/provider/shared/utils.go
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-package provider
+package shared
 
 import (
 	"net"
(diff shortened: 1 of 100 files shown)


internal/provider/shared/utils.go
```diff
@@ -15,45 +15,43 @@ import (
 	"github.com/oxidecomputer/oxide.go/oxide"
 )
 
-// replaceBackticks replaces ” with `. It can be used to defined codeblocks in
+// ReplaceBackticks replaces " with `. It can be used to defined codeblocks in
```
sudomateo: This should change back to the smart quote because `go fmt` turns `''` into `”`. 

```suggestion
// ReplaceBackticks replaces ” with `. It can be used to defined codeblocks in
```

lgfa29 (approved):
LGTM! 

Kind of unfortunate that git lost track of history from `testutils.go` to `sharedtest.go`, but git works in mysterious ways, and this is not a very important file to keep track of changes.

sudomateo:
Thanks @lgfa29! Going to merge this after we release the R19 compatible release of Terraform.

lgfa29 (approved):

PR #745, opened 2026-04-14
ci: enable `goimports` lint

lgfa29:
Lint `import` blocks to follow `goimports` rules.

Also remove depracted `golangci.yaml` file.


jmcarp (approved):
Nice, 🚢

PR #746, opened 2026-04-16
Bump oxide.go and use errors.Is.

jmcarp:
Use improved error handling in oxide.go to check structured error codes rather than substring matches.

This turned out to be a minimal change because we really only care about 404s here so far, and already have a wrapper for that check.


sudomateo:
Closing this. We will switch to the error types as we work on the provider.

PR #747, opened 2026-04-16
misc: explicit resource and data source model types

sudomateo:
Updated the model types to be explicit for each resource and data source. Since a given resource and data source now live in the same package, they can conflict on model type names, though it's unlikely. In fact, a case could be made to consolidate models for a resource and data source but that's not a concern for this patch.

This patch ensures the model type for a resource is named `ResourceModel` and the model type for a data source is named `DataSourceModel`. Then, each field for a model that uses a struct type is named with a `ResourceModel` or `DataSourceModel` suffix to show that the field belongs to the resource or data source.

Relates to SSE-266.

### DIFF
--- a/internal/provider/address_lot/datasource.go
+++ b/internal/provider/address_lot/datasource.go
@@ -26,7 +26,7 @@ type DataSource struct {
 }
 
 type DataSourceModel struct {
-	Blocks       []DataSourceBlockModel `tfsdk:"blocks"`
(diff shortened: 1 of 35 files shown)


lgfa29 (approved):
This reads much better, thanks!

Similar to #748, I would like to wait until CI is fixed to merge, just in case.

lgfa29:
Rebasing to get #750

PR #750, opened 2026-04-20
ci: update and fix tests

lgfa29:
Fix acceptance tests to handle the new project structure.
  - Test all packages by default instead of just `internal`.
  - Collect `terraform.log` files from all packages.
  - Fix subnet assignment to isolate subnet generation per test. The shared `NextSubnetCIDR()` helper now has new state per package.
  - Remove Python dependency cache warning.

Fix and update unit tests.
  - Download Terraform explicitly instead of relying on auto-download. This removes failures in case the Terraform Framework can't find Terraform or if Checkpoint is down. Currently tests are failing with the error `failed to find or install Terraform CLI from [0xc00024e100 0xc00016f560]: unable to verify checksums signature: openpgp: key expired`, maybe because of https://discuss.hashicorp.com/t/hcsec-2026-03-hashicorp-gpg-key-72d7468f-update/77237.
  - Test using OpenTofu.
  - Upload results to TestRail.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ endif
 
 # Acceptance test variables
 TEST_ACC_COUNT ?= 1
-TEST_ACC ?= github.com/oxidecomputer/terraform-provider-oxide/internal/provider
+TEST_ACC ?= github.com/oxidecomputer/terraform-provider-oxide/internal/...
 TEST_ACC_NAME ?= TestAcc
 TEST_ACC_PARALLEL = 6
 TEST_ACC_OMICRON_BRANCH ?=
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -18,7 +18,7 @@ endif
 
 # Acceptance test variables
 TEST_ACC_COUNT ?= 1
-TEST_ACC ?= github.com/oxidecomputer/terraform-provider-oxide/internal/provider
```
sudomateo, line 21: Sorry about this!

sudomateo (approved):

Makefile
lgfa29: No worries! I'm more concerned about CI passing when all tests are skipped 😅 

But I will investigate this later.

PR #751, opened 2026-04-22
Update provider attribute docs to match openapi.

jmcarp:
Our resource attribute descriptions are generally very similar to the corresponding field descriptions in openapi. Sometimes, these descriptions diverge over time. This patch reconciles attribute descriptions for the instance resource, and adds a skill to reconcile descriptions in the future. We propose this as a skill for an agent rather than a deterministic program because descriptions *should* diverge between provider and openapi when behavior differs, or when we can't map provider to openapi 1:1.

-----

### Pull request checklist

### DIFF
--- a/.agents/skills/sync-schema-descriptions/SKILL.md
+++ b/.agents/skills/sync-schema-descriptions/SKILL.md
@@ -1,0 +1,208 @@
+---
+name: sync-schema-descriptions
+description: |
+  Sources Terraform schema descriptions from the pinned oxide.go SDK's Go doc
+  comments, which already mirror the OpenAPI spec. Adapts the text to the TF
+  surface (translating reshaped attribute names, clarifying NameOrId inputs,
+  trimming API-internal paragraphs, and reconciling any divergence between
+  provider and API default behavior) so the descriptions stay accurate and
+  aligned with the version of the SDK the provider actually calls into.
+---
+
+# Sync Schema Descriptions
+
+## Overview
+
+Terraform schema `Description` / `MarkdownDescription` strings are hand-written
+and drift from the canonical descriptions in the OpenAPI spec. The oxide.go
+SDK, pinned via `go.mod`, has the same descriptions baked into Go doc comments
+on each field. This skill walks through sourcing the right description,
+adapting it, and applying the change.
+
+Not every TF attribute maps 1:1 to a single SDK field (e.g. TF's `boot_disk_id`
+corresponds to the SDK's richer `boot_disk`), so blind copying is wrong. The
+skill codifies the adaptation process rather than automating it blindly.
+
+## Workflow
+
+### 1. Resolve the SDK source
+
+    go list -m -f '{{.Dir}}' github.com/oxidecomputer/oxide.go
+
+Descriptions live in `<dir>/oxide/types.go`. Do not use a local checkout of
+the SDK, which might not match the pinned version.
+
+### 2. Identify scope
+
+The user will usually ask about one attribute (e.g. `instance.ssh_public_keys`),
+a whole resource, or all resources. For each in-scope attribute, work through
+steps 3–8.
+
+### 3. Locate the TF attribute
+
+In `internal/provider/<resource>/resource.go` (or `data_source.go`), find the
+schema block and note the current `Description` / `MarkdownDescription`. Find
+the TF model struct field via its `tfsdk:"<attr_name>"` tag.
+
+Do not touch schema-version-compat files (`resource_v0.go`, `resource_v1.go`,
+…). They represent frozen prior schemas kept for state upgrades; descriptions
+there are not rendered to current docs, so the files should stay as they
+were when that version shipped.
+
+### 4. Find the matching SDK field
+
+**Input attributes (set in Create/Update):** grep the Create method for the
+model field. You'll see a line like `params.Body.SshPublicKeys = ...`. The SDK
+struct is the type of `params.Body` (e.g. `oxide.InstanceCreate`); the SDK
+field is the LHS accessor.
+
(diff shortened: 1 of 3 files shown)


internal/provider/instance/resource.go
```diff
@@ -311,12 +311,12 @@ This resource manages instances.
 			},
 			"disk_attachments": schema.SetAttribute{
 				Optional:            true,
-				MarkdownDescription: "IDs of the disks to be attached to the instance. When multiple disk IDs are provided, set `boot_disk_id` to specify the boot disk for the instance. Otherwise, a boot disk will be chosen randomly.",
+				MarkdownDescription: "Names or IDs of the disks to be attached to the instance. The order of this list does not guarantee a boot order for the instance. When `boot_disk_id` is specified it will count against the disk attachment limit.",
```
lgfa29: > When `boot_disk_id` is specified it will count against the disk attachment limit.

This new sentence sounds a little weird to me. Does this mean the `boot_disk_id` will count twice since it's required to be added to this list? I think this would make sense if we stopped requiring the `boot_disk_id` to be add here, but that's not the case right now.

internal/provider/instance/resource.go
```diff
@@ -288,12 +288,12 @@ This resource manages instances.
 			},
 			"anti_affinity_groups": schema.SetAttribute{
 				Optional:    true,
-				Description: "IDs of the anti-affinity groups this instance should belong to.",
+				Description: "Names or IDs of the anti-affinity groups this instance should belong to.",
```
lgfa29: I guess that's the case, but out of curiosity, have you checked if this is actually true? I think I (and our test suite) usually only exercise passing IDs.

internal/provider/instance/resource_v0.go
lgfa29, line 1: It shouldn't really matter, but I would prefer to keep these files (v0 and v1) intact since they're kind of a snapshot of the resource schema from the time the migration was created.

internal/provider/instance/resource.go
jmcarp: Yep, I actually noticed this error in the docs while working on the CAPI provider.

internal/provider/instance/resource_v0.go
jmcarp: Reverted.

internal/provider/instance/resource.go
jmcarp: I take it back! NameOrId "works", except that it also causes infinite spurious diff/recreate on subsequent applies. Wrote this up in #752, but I'm going to update the skill and generated docs to use ids only for now.

.agents/skills/adapt-api-desc/SKILL.md
```diff
@@ -0,0 +1,171 @@
...
+
+### 3. Locate the TF attribute
+
+In `internal/provider/<resource>/resource.go` (or `data_source.go`), find the
+schema block and note the current `Description` / `MarkdownDescription`. Find
+the TF model struct field via its `tfsdk:"<attr_name>"` tag.
+
+If the resource has schema-version-compat files (`resource_v0.go`,
+`resource_v1.go`, …), the same description is duplicated there; plan to update
+all copies.
```
lgfa29: I think we don't want to update all copies?
https://github.com/oxidecomputer/terraform-provider-oxide/pull/751#discussion_r3127212889

.agents/skills/adapt-api-desc/SKILL.md
```diff
@@ -0,0 +1,171 @@
...
+needs a hand-written description, which is outside this skill's scope.
+Surface the gap to the operator. Never copy text from a TF sibling
+(`v4` → `v6`, `ephemeral` → `floating`, etc.) and pass it off as adapted; it
+is still hand-written, just plagiarized.
+
+**a. Translate field-name references.** If the comment references an API field
+name that differs in TF (e.g. "set the `boot_disk` attribute"), rewrite it to
+the TF name (`boot_disk_id`). Same for any other cross-referenced field.
+
+**b. `NameOrId` phrasing — use "ID", not "name or ID".** The API accepts both
```
lgfa29: Does this need to be "name OR ID"? For example, this attribute should still use name, even though I believe this is `NameOrId` in the SDK.

https://github.com/oxidecomputer/terraform-provider-oxide/blob/18a8c8700a91f40bd9e50548b73a48b3b3e49b02/internal/provider/project/datasource.go#L75

.agents/skills/sync-schema-descriptions/SKILL.md
```diff
@@ -0,0 +1,171 @@
...
+- information already encoded by the TF schema (e.g. defaults — TF renders
+  those separately)
+
+Keep paragraphs describing user-visible behavior: accepted value forms,
+interactions with other attributes. Paragraphs about omission/defaults need
+verification first — see rule (e).
+
+**d. Preserve TF-specific additions.** If the current TF description contains
+information that isn't in the SDK (e.g. cross-refs to other TF attrs, validator
+behavior), keep it — it usually reflects provider logic that doesn't exist at
```
lgfa29, line 130: In addition to keep existing TF-specifc information, should the agent look to _update_ things as well? For example, fields that have validators that are currently undocumented.

lgfa29 (approved):
Some minor comments, mostly double-checking what the agent would/should do in some cases.

.agents/skills/adapt-api-desc/SKILL.md
jmcarp: Yep, updated the skill.

jmcarp:
I want to do more thinking about when this runs and whether we want to apply it to more resources, but I'm happy with this as a first pass, and I believe all comments are addressed. Merging.

PR #753, opened 2026-04-23
Test examples.

### LINKED ISSUE #392: Update examples to cover common use cases

jmcarp:
Our examples are useful, but become stale over time. This patch treats each example config file as an acceptance test, verifying that we can apply the config without errors. We also make changes to cause some of the examples to become applicable, since many aren't: pre-creating shared resources in scripts/acc-test-setup.sh, referring to resources using data sources rather than hard-coded ids, and fixing occasional bugs in the provider or examples.

-----

### Pull request checklist


docs/resources/anti_affinity_group.md
```diff
@@ -25,6 +25,11 @@ resource "oxide_anti_affinity_group" "example" {
     update = "2m"
   }
 }
+
+# Prerequisites for the example.
```
lgfa29, line 29: This comment doesn't seem relevant to users. We can probably leave it out.

internal/provider/floating_ip/datasource.go
```diff
@@ -164,10 +164,6 @@ func (f *DataSource) Read(
 
 	floatingIP, err := f.client.FloatingIpView(ctx, params)
 	if err != nil {
-		if shared.Is404(err) {
-			resp.State.RemoveResource(ctx)
-			return
-		}
```
lgfa29, line 170: Why does this needs to be removed?

internal/provider/examples_test.go
```diff
@@ -0,0 +1,87 @@
...
+				t.Skip()
+			}
+
+			// Subtests run serially: several examples attach the shared
+			// seeded disk, and a disk can only be attached to one
+			// instance at a time.
+			hcl, err := os.ReadFile(path)
+			if err != nil {
+				t.Fatalf("read %s: %v", path, err)
+			}
```
lgfa29, line 71:
```suggestion
			require.NoError(t, err)
```

The other tests use testify, so better to keep things consistent.

internal/provider/examples_test.go
```diff
@@ -0,0 +1,87 @@
...
+					{Config: string(hcl)},
+				},
+			})
+		})
+
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("walk %s: %v", examplesRoot, err)
+	}
```
lgfa29, line 86:
```suggestion
	require.NoError(t, err)
```

Same here.

lgfa29 (approved):
Great addition! I just left some minor comments.

internal/provider/floating_ip/datasource.go
jmcarp: This is a provider bug that the example tests revealed, but I'll back it out here and write it up in a separate patch so it's easier to review.

docs/resources/anti_affinity_group.md
jmcarp: Done.

PR #757, opened 2026-05-08
test: add TEST_ACC_SIM_HOST variable

lgfa29:
Add the `TEST_ACC_SIM_HOST` variable to `Makefile` to allow running some targets against a remote simulated omicron. This is useful in scenarios where the simulator and the tests are running in different machines.


sudomateo (approved):
EZPZ. I remember manually modifying this when I ran against my remote Helios machine. Thanks for adding this.

PR #758, opened 2026-05-09
docs: add upgrade guide page

lgfa29:
Add documentation page detailing how to deal with deprecations and breaking changes when upgrading the provider version. This page must be updated every time a release contains a deprecation or a breaking change.

I only went back as far as `0.16.0` because I think that's the earliest relevant version for us today. We can add later versions if necessary in the future.

### DIFF
--- a/docs/guides/upgrade.md
+++ b/docs/guides/upgrade.md
@@ -1,0 +1,353 @@
+---
+page_title: "Oxide Provider - Version Upgrade Guide"
+description: |-
+  Instructions on how to upgrade the Oxide provider to versions that contain
+  deprecations and breaking changes.
+---
+
+# Oxide Provider Version Upgrade Guide
+
+This page documents additional steps and instructions that may be necessary
+when upgrading the Oxide Terraform provider to a given version.
+
+It may be necessary to apply all version changes when upgrading multiple
+versions at a time.
+
+Refer to the
+[changelog](https://github.com/oxidecomputer/terraform-provider-oxide/blob/main/CHANGELOG.md)
+for a full list of changes.
+
+## Upgrading to `0.19.0`
+
+Release `0.19.0` contains breaking changes that require updates to Terraform
+configuration files.
+
+### Resource `oxide_instance`
+
+#### Breaking change: removed deprecated attributes
+
+The deprecated attributes `host_name`, `network_interfaces.ip_address`,
+`network_intefaces.mac_address`, `network_interfaces.id`,
+`network_interfaces.primary`, `network_interfaces.time_created`, and
+`network_interfaces.time_modified` have been removed. Use `hostname` and the
+`attached_network_interfaces` instead.
+
+Refer to previous deprecation notes for more details on to how to update
+configuration files.
+
+### Function `to_vpc_firewall_rules_map`
+
+#### Breaking change: removed `to_vpc_firewall_rules_map` function
+
+The `to_vpc_firewall_rules_map` function has been removed.
+
+Refer to previous breaking change notes of `vpc_firewall_rules` for more
+details on to how to update configuration files.
+
+## Upgrading to `0.18.0`
+
+Release `0.18.0` contains breaking changes that require updates to Terraform
+configuration files and deprecations that may require changes when upgrading to
+future versions.
+
+### Resource `oxide_instance`
+
+#### Breaking change: changed schema of `external_ips`
+
+The `external_ips` attribute of the `oxide_instance` resource is now an object
+instead of a list. The new schema adds support for ephemeral IPv6.
+`oxide_instance` resources need to be updated to match the new schema.
+
+Before:
+
+```terraform
+resource "oxide_instance" "instance" {
+  # ...
+  external_ips = [
+    # External IP from the default pool.
+    {
+      type = "ephemeral"
+    },
+
+    # External IP from a specific IP pool.
+    {
+      type = "ephemeral"
+      id   = "4f0e69ad-66b6-41c0-b727-7b0285b0c384"
+    },
+
+    # Floating IP.
+    {
+      type = "floating"
+      id   = "eb65d5cb-d8c5-4eae-bcf3-a0e89a633042"
+    }
+  ]
+  # ...
+}
+```
+
+After:
+
+```terraform
+resource "oxide_instance" "instance" {
+  # ...
+  external_ips = {
+    ephemeral = [
+      # External IPv4 from the default pool.
+      {
+        ip_version = "v4"
+      },
+
+      # External IP from a specific IP pool.
+      {
+        ip_version = "v4"
+        pool_id    = "4f0e69ad-66b6-41c0-b727-7b0285b0c384"
+      },
+    ]
+
+    floating = [
+      # Floating IP.
+      {
+        id = "eb65d5cb-d8c5-4eae-bcf3-a0e89a633042"
+      },
+    ]
+  }
+  # ...
+}
+```
+
+#### Deprecation: `network_interfaces.ip_address`
+
+The `ip_address` attribute of `network_interfaces` elements has been deprecated
+and will removed in a future release. Use `ip_config` instead.
+
+The new `ip_config` attribute allows using IPv6 addresses.
+
+Before:
+
+```terraform
+resource "oxide_instance" "instance" {
+  # ...
+  network_interfaces = [
+    # Network interface with auto-assigned IP.
+    {
+      name        = "nic0"
+      description = "nic0"
+      vpc_id      = data.oxide_vpc_subnet.default.vpc_id
+      subnet_id   = data.oxide_vpc_subnet.default.id
+    },
+
+    # Network interface with specific IP.
+    {
+      name        = "nic1"
+      description = "nic1"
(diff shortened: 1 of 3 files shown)


templates/guides/upgrade.md.tmpl
sudomateo, line 1: We're not actually rendering anything in this template, right? We're just using it so `make generate` will turn the template into markdown?

docs/guides/upgrade.md
```diff
@@ -0,0 +1,353 @@
...
+
+#### Deprecation: computed attributes of `network_interfaces` elements
+
+The computed attributes `id`, `mac_address`, `primary`, `time_created`, and
+`time_modified` attributes of `network_interfaces` elements have been
+deprecated and will be removed in a future release. Use the new
+`attached_network_interfaces` attribute to access these values.
+
+Network interfaces in the new `attached_network_interfaces` attribute can be
+reference by name instead of having to iterate over a set.
```
sudomateo, line 198:
```suggestion
referenced by name instead of having to iterate over a set.
```

docs/guides/upgrade.md
```diff
@@ -0,0 +1,353 @@
...
+
+```terraform
+resource "oxide_instance" "instance" {
+  # ...
+  hostname = "my-instance"
+  # ...
+}
+```
+
+## Upgrading to `0.16.0`
```
sudomateo, line 285: What do you think about listing each version, regardless whether it contains special instructions?

```suggestion
## Upgrading to `0.17.0`

There are no special considerations when upgrading to this release.

## Upgrading to `0.16.0`
```

sudomateo (approved):
I think this looks good overall. I'm unsure what name will show up in the sidebar in the Terraform Registry, but we'll change it if we don't like it.

I left some feedback and a typo fix but I'm approving so I don't block.

hermanol:
The content as written looks great, and addresses all the points from the CS issue for 17 -> 18 -> 19. 

I did have three thoughts on formatting/process for keeping the guide current and accessible going forward, none of which are immediate concerns:

1. Would it make sense to have individual advisements under each release be [collapsible](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections)? This would probably be necessary only if we have 3+ versions all on one page, but the code blocks do take up a lot of vertical space. 

2. I noticed the template inclusion as well, I think that that could be useful in a possible workflow where we have several version advisories up at a time, which would then be archived and put into a `past-releases.md` file in the docs folder past a certain expiry date. That being said, there’s a lot of details in there from this batch that could be stripped down to headers/formatting, if the intention was to re-use it for other batches of releases.

3. On the [docs site](https://docs.oxide.computer/guides/integrations/terraform) we link out to the Terraform Provider Registry page [here](https://registry.terraform.io/providers/oxidecomputer/oxide/latest/docs). Would it make sense to link out to include a quick line there linking out to these release-to-release changes as well? I figure most operators will be coming in through Github, but if they were to come in through the docs site, they'd probably hit the Terraform registry page first.

hermanol (approved):

sudomateo:
We could take the [approach other providers do](https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/version_2_upgrade) and have a "Guides" sidebar with a new document for each release.

The Terraform Registry renders heading links for each document so it's probably not necessary to have collapsible sections since a user can just click the heading they want to read.

templates/guides/upgrade.md.tmpl
lgfa29: Yup. It needs to be in the `templates` folder otherwise `tfplugindocs` will delete it from `docs`. It's currently not a template but I was trying to have each version in a separate file and load them in this page, but I couldn't get that to work. 

We could use some of the `tfplugindocs` template functionality, like loading external `*.tf` files, but I think in this case it would make things harder to manage?

I just tested and we don't actually need `tmpl` extension, so I can also remove that.

docs/guides/upgrade.md
lgfa29: I thought about it, but I ended going against because:
1. It would add something else for us to do during a release. Not a big deal, but just one more thing to do and track.
2. The page is already a bit long, and I'm not sure how much value these would add, but maybe it could confusing to not have some versions?

lgfa29:
> The content as written looks great, and addresses all the points from the CS issue for 17 -> 18 -> 19.
> 
> I did have three thoughts on formatting/process for keeping the guide current and accessible going forward, none of which are immediate concerns:
> 
> 1. Would it make sense to have individual advisements under each release be [collapsible](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections)? This would probably be necessary only if we have 3+ versions all on one page, but the code blocks do take up a lot of vertical space.

This could be useful. I thought the Terraform registry website didn't support them, but maybe it does? You can use https://registry.terraform.io/tools/doc-preview to preview how markdown files are going to be rendered in the registry.

My only caveat is that it may be easy to miss the collapsible sections among the page content?

> 2. I noticed the template inclusion as well, I think that that could be useful in a possible workflow where we have several version advisories up at a time, which would then be archived and put into a `past-releases.md` file in the docs folder past a certain expiry date. That being said, there’s a lot of details in there from this batch that could be stripped down to headers/formatting, if the intention was to re-use it for other batches of releases.

I don't quite follow this part:
> That being said, there’s a lot of details in there from this batch that could be stripped down to headers/formatting, if the intention was to re-use it for other batches of releases.

Would you suggest a different layout for this page? Currently the headers are organized as

```
+- Version
+--- Resource
+----- Change
```

Another approach I thought was to have change type first. Something like:

```markdown
## Upgrading to `0.19.0`
...
### Breaking Changes
#### Resource `oxide_instance`
##### RemovedX
...
### Deprecations
#### Resource `oxide_image`
##### Deprecated X
...
```

> 3. On the [docs site](https://docs.oxide.computer/guides/integrations/terraform) we link out to the Terraform Provider Registry page [here](https://registry.terraform.io/providers/oxidecomputer/oxide/latest/docs). Would it make sense to link out to include a quick line there linking out to these release-to-release changes as well? I figure most operators will be coming in through Github, but if they were to come in through the docs site, they'd probably hit the Terraform registry page first.

This doc will be include in the registry page under a `Guides` section. Check https://registry.terraform.io/providers/hashicorp/google/latest/docs for an example.

lgfa29:
> We could take the [approach other providers do](https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/version_2_upgrade) and have a "Guides" sidebar with a new document for each release.

I have a vague memory of us deciding against having separate pages, but I don't actually recall the reasoning so I may be misremembering things. I fine going either way.

> The Terraform Registry renders heading links for each document so it's probably not necessary to have collapsible sections since a user can just click the heading they want to read.

I should've included this in the PR body, but you can see how it will look like using the Docs Preview tool I linked earlier.
image

templates/guides/upgrade.md.tmpl
sudomateo: No worries at all. I just wanted to confirm. The extension is fine with me.

docs/guides/upgrade.md
sudomateo: That works for me. I looked at other upgrade pages and the existing format is consistent with those pages. I'm cool to keep it as-is.

sudomateo:
> I should've included this in the PR body, but you can see how it will look like using the Docs Preview tool I linked earlier.

Love the way this looks! I can also go either way. It looks like we just need to make a directory for the Terraform Registry to render things in a collapsible sidebar section. At least that's the idea I've seen here: https://github.com/hashicorp/terraform-provider-google/tree/main/website/docs/guides

lgfa29:
> It looks like we just need to make a directory for the Terraform Registry to render things in a collapsible sidebar section. At least that's the idea I've seen here: https://github.com/hashicorp/terraform-provider-google/tree/main/website/docs/guides

Yup, each markdown file under `docs/guides` becomes its own page under the `Guides` section. There are also subcategories, but I've never used them, so I'm not sure how they will look like. It's a little unfortunate that the docs preview tool doesn't render them.

https://developer.hashicorp.com/terraform/registry/providers/docs#guides

lgfa29:
> Would you suggest a different layout for this page? Currently the headers are organized as
> 
> ```
> +- Version
> +--- Resource
> +----- Change
> ```
> 
> Another approach I thought was to have change type first. Something like:
> 
> ```
> ## Upgrading to `0.19.0`
> ...
> ### Breaking Changes
> #### Resource `oxide_instance`
> ##### RemovedX
> ...
> ### Deprecations
> #### Resource `oxide_image`
> ##### Deprecated X
> ...
> ```

I prototyped this approach and didn't quite like it was rendered.

Headers after `<h4>` are rendered the same, so it looks a little weird.

image

I also tried to prefix the last header, but it looks a little repetitive when there are multiple entries for the same resource.

image

sudomateo:
I see what you mean. I don't like the way that looks either. If we did move these pages under some "Upgrade Guides" sidebar section then we can reclaim a heading level for better use. Not saying that we should do that but perhaps it's the better middle ground in terms of organization and rendering style.

PR #766, opened 2026-06-10
Load `silo_saml_identity_provider` resource into state if it already exists

### LINKED ISSUE #761: `oxide_silo_saml_identity_provider`: restore state on create
The `oxide_silo_saml_identity_provider` resource is an unusual Terraform resource because it doesn't have endpoints to delete or update it. On deletion, the resource is removed from state but it still exists in the backend. This creates a problem because it's not impossible to `terraform apply` a configuration that declares this resource.

It would be possible to `terraform apply` after removing the resource from the configuration, but this is not always feasible. For example, if the resource is part of a module that is used multiple times, removing the `oxide_silo_saml_identity_provider` would cause it to be removed from every module instance.

It's also currently not possible to import a `oxide_silo_saml_identity_provider` due to the asymmetric API for create and read.

Since this kind of resource cannot be deleted, the `Create()` method should restore it to state if the IdP configuration already exists, reading the values not present in the `GET` request (like `idp_metadata_source`) from the user configuration file.

This could cause some inconsistency between the Terraform state and the backend if the asymmetric attributes differ (for example, if the user sets a different value for  `idp_metadata_source` after it is deleted from state), but since this resource can't be updated, the different value will only be a problem if users are relying on it as input for something else, which seems unlikely given how specific these attributes are. We should consider marking attributes not returned on `GET` as write-only to avoid this problem.

bwagner5:
Fixes https://github.com/oxidecomputer/terraform-provider-oxide/issues/761

The silo_saml_identity_provider resource cannot be deleted once created. It can be removed from Terraform state, but adding it back was problematic since it would already exist in the Oxide control plane. This change handles the "already exists" error in the Create path which triggers a read of the resource to update Terraform's state. 

Fixed a few test issues related error propagation and formatting. 

-----

### Pull request checklist

### DIFF
--- a/.changelog/0.20.0.toml
+++ b/.changelog/0.20.0.toml
@@ -1,0 +1,3 @@
+[[enhancements]]
+title = "support adoption for the silo_saml_identity_provider"
+description = "the silo_saml_identity_provider (create-only resource) can now be removed from tf state and added back without issue [#766](https://github.com/oxidecomputer/terraform-provider-oxide/pull/766)."
--- a/.changelog/0.20.0.toml
+++ b/.changelog/0.20.0.toml
@@ -1,3 +1,3 @@
 [[enhancements]]
-title = "support adoption for the silo_saml_identity_provider"
+title = "Load `silo_saml_identity_provider` resource into state if it already exists"
 description = "the silo_saml_identity_provider (create-only resource) can now be removed from tf state and added back without issue [#766](https://github.com/oxidecomputer/terraform-provider-oxide/pull/766)."
--- a/.changelog/0.20.0.toml
+++ b/.changelog/0.20.0.toml
@@ -1,3 +1,3 @@
 [[enhancements]]
 title = "Load `silo_saml_identity_provider` resource into state if it already exists"
-description = "the silo_saml_identity_provider (create-only resource) can now be removed from tf state and added back without issue [#766](https://github.com/oxidecomputer/terraform-provider-oxide/pull/766)."
+description = "The `silo_saml_identity_provider` (create-only resource) can now be removed from state and added back without issue [#766](https://github.com/oxidecomputer/terraform-provider-oxide/pull/766)."
--- a/internal/provider/silo_saml_identity_provider/resource.go
+++ b/internal/provider/silo_saml_identity_provider/resource.go
@@ -391,7 +391,8 @@ func (r *Resource) Read(
 	)
 
 	// GroupAttributeName is Optional so we need to handle the case where it is not set in the
-	// remote Oxide control plane and take whatever value the configuration has (Null or empty string)
+	// remote Oxide control plane and take whatever value the configuration has (Null or empty
+	// string)
 	if idpConfig.GroupAttributeName != "" {
 		state.GroupAttributeName = types.StringValue(idpConfig.GroupAttributeName)
 	}
@@ -469,13 +470,21 @@ func (m ResourceModel) Diff(remote *oxide.SamlIdentityProvider) diag.Diagnostics
 	}{
 		{path.Root("acs_url"), m.AcsUrl.ValueString(), remote.AcsUrl},
 		{path.Root("description"), m.Description.ValueString(), remote.Description},
-		{path.Root("group_attribute_name"), m.GroupAttributeName.ValueString(), remote.GroupAttributeName},
+		{
+			path.Root("group_attribute_name"),
+			m.GroupAttributeName.ValueString(),
+			remote.GroupAttributeName,
+		},
 		{path.Root("idp_entity_id"), m.IdpEntityId.ValueString(), remote.IdpEntityId},
 		{path.Root("name"), m.Name.ValueString(), string(remote.Name)},
 		{path.Root("signing_keypair").AtName("public_cert"), signingPublicCert, remote.PublicCert},
 		{path.Root("slo_url"), m.SloUrl.ValueString(), remote.SloUrl},
 		{path.Root("sp_client_id"), m.SpClientId.ValueString(), remote.SpClientId},
-		{path.Root("technical_contact_email"), m.TechnicalContactEmail.ValueString(), remote.TechnicalContactEmail},
+		{
+			path.Root("technical_contact_email"),
+			m.TechnicalContactEmail.ValueString(),
+			remote.TechnicalContactEmail,
+		},
 	}
 
 	var diagnostics diag.Diagnostics
--- a/internal/provider/silo_saml_identity_provider/resource.go
+++ b/internal/provider/silo_saml_identity_provider/resource.go
@@ -304,37 +304,38 @@ func (r *Resource) Create(
 		// Diff the remote state with the configuration before adopting the resource
 		// If any parameters differ, diagnostics are returned and we should abort to display
 		// errors to the user.
-		diagnostics := plan.Diff(idpConfig)
-		if diagnostics.HasError() {
+		if diagnostics := plan.Diff(idpConfig); diagnostics.HasError() {
 			resp.Diagnostics.Append(diagnostics...)
 			return
 		}
 
 		// group_attribute_name is Optional. Diff treats null and "" as
-		// equivalent, but state must match the canonical representation
-		// Terraform's plan layer uses for an Optional StringAttribute —
-		// otherwise the next refresh-plan reports a phantom drift between
-		// state ("") and config (null). When the remote is empty, force the
-		// plan field to StringNull() regardless of which empty form the user
-		// wrote in config.
+		// equivalent, but state must match between the remote state and configuration.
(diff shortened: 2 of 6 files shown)


.changelog/0.20.0.toml
```diff
@@ -0,0 +1,3 @@
+[[enhancements]]
+title = "support adoption for the silo_saml_identity_provider"
```
lgfa29:
```suggestion
title = "Load `silo_saml_identity_provider` resource into state if it already exists"
```

A bit more verbose, but I think it's a little more clear than "support adoption". It also links the changelog to the error an user would see before.

.changelog/0.20.0.toml
```diff
@@ -0,0 +1,3 @@
...
+description = "the silo_saml_identity_provider (create-only resource) can now be removed from tf state and added back without issue [#766](https://github.com/oxidecomputer/terraform-provider-oxide/pull/766)."
```
lgfa29:
```suggestion
description = "The `silo_saml_identity_provider` (create-only resource) can now be removed from state and added back without issue [#766](https://github.com/oxidecomputer/terraform-provider-oxide/pull/766)."
```

Minor suggestions.

internal/provider/silo_saml_identity_provider/resource.go
```diff
@@ -277,7 +278,35 @@ func (r *Resource) Create(
 	}
 
 	idpConfig, err := r.client.SamlIdentityProviderCreate(ctx, params)
-	if err != nil {
+	// if the IDP already exists, then try to adopt it
+	if errors.Is(err, oxide.ErrObjectAlreadyExists) {
```
lgfa29, line 283: @jmcarp will be happy to see this being used 😄

internal/provider/silo_saml_identity_provider/resource.go
```diff
@@ -277,7 +278,35 @@ func (r *Resource) Create(
...
+			ctx,
+			oxide.SamlIdentityProviderViewParams{
+				Silo:     oxide.NameOrId(plan.Silo.ValueString()),
+				Provider: oxide.NameOrId(plan.Name.ValueString()),
+			},
+		)
+		if viewErr != nil {
+			resp.Diagnostics.AddError(
+				"Error adopting SAML identity provider",
+				fmt.Sprintf("Create API error: %s, View API error: %s", err, viewErr),
```
lgfa29: Maybe hard to test, but have you had a chance to see how this error is displayed to users? I have a felling the text will be formatted quite weirdly because `err` and `viewErr` likely have line breaks.

It's probably fine to ignore `err` here since we know `errors.Is(err, oxide.ErrObjectAlreadyExists)` is `true`?

internal/provider/silo_saml_identity_provider/resource.go
```diff
@@ -277,7 +278,35 @@ func (r *Resource) Create(
...
+			)
+			return
+		}
+		plan.AcsUrl = types.StringValue(idpConfig.AcsUrl)
```
lgfa29: Maybe worth adding a comment explaining that some attributes are not returned by the API on read, so they're retained from the plan, which is whatever the user provided in their config.

internal/provider/silo_saml_identity_provider/resource.go
```diff
@@ -277,7 +278,35 @@ func (r *Resource) Create(
...
+		plan.Description = types.StringValue(idpConfig.Description)
+		plan.GroupAttributeName = types.StringValue(idpConfig.GroupAttributeName)
```
lgfa29: I think we need to check for an empty string because this attribute is optional, so if not provided in the config we would be changing the plan form a `Null` string to an empty string. Something like:

```suggestion
		if idpConfig.GroupAttributeName != "" {
				plan.GroupAttributeName = types.StringValue(idpConfig.GroupAttributeName)
		}
```

One way to test this would be to not set this value in the test and see what we get back. We should see a `Null` rather than `""`. The impact of this, I think, is that `terraform plan` would try to do an update on this resource.

I think that's the only `Optional` attribute we care about, but worth double-checking.

internal/provider/silo_saml_identity_provider/resource_test.go
```diff
@@ -29,6 +29,7 @@ type resourceConfig struct {
 	SiloName                          string
 	SiloSamlIdentityProviderBlockName string
 	SiloSamlIdentityProviderName      string
+	RemoveSamlIdentityProvider        bool
```
lgfa29: I usually avoid double-negatives (`RemoveSamlIdentityProvider: false` means the SAML config is present, which is a positive outcome), but I see how it makes sense to have `false` as the useful default in this case.

A minor suggestion would be name this something like `SkipSamlIdentityProvider`, which is a positive action, so `SkipSamlIdentityProvider: false` doesn't sound like a double-negative.

lgfa29 (changes requested):
Some minor changes, but overall this LGTM!

internal/provider/silo_saml_identity_provider/resource.go
bwagner5: it's very nice! thanks @jmcarp !

internal/provider/silo_saml_identity_provider/resource.go
bwagner5: I removed the ObjectAlreadyExists error so it's not showing two since the other one is expected. I used the already exists error to test the output though since that's easier than trying to simulate a read failure. 

```
│ Error: Error adopting SAML identity provider
│
│   with oxide_silo_saml_identity_provider.idp,
│   on main.tf line 70, in resource "oxide_silo_saml_identity_provider" "idp":
│   70: resource "oxide_silo_saml_identity_provider" "idp" {
│
│ View API error: POST http://localhost:12220/v1/system/identity-providers/saml?silo=cf5b74e7-04db-487a-9c1d-ed546cf5d562
│ ----------- RESPONSE -----------
│ Status: 400 ObjectAlreadyExists
│ Message: already exists: saml-identity-provider "saml-test"
│ RequestID: aa87eb6b-c1d5-4393-8dad-804dacdf49a4
│ ------- RESPONSE HEADERS -------
│ Date: [Fri, 12 Jun 2026 01:54:00 GMT]
│ Content-Type: [application/json]
│ X-Request-Id: [aa87eb6b-c1d5-4393-8dad-804dacdf49a4]
│ Content-Length: [166]
│
```

lgfa29: One thing I forgot to mention is that we can add as many diagnostics as we want before returning, so we could have each error be a separate message, but I think that would be confusing in this case since we probably only care about the inner error.

bwagner5: agreed, the other error is more confusing anyways since it's a supported path and something expected in that case.

internal/provider/silo_saml_identity_provider/resource.go
bwagner5: good catch! I had to change a few other areas to make this work and protect against a Null on initial creation and then a change to the optional field on an adoption.

lgfa29: > protect against a Null on initial creation and then a change to the optional field on an adoption.

Oh right, we can't have a diff here because it's a create operation, so there Terraform will complain about an inconsistent apply. And this is a very confusing error because Terraform will try to blame the provider 😅 

To make things worse, this is true for all attributes, not just optional ones, so I think we'll need to apply similar logic to all of them, and plus cover the case where the attribute is set in both the config and the remote state, but they're different.

I would suggest adding a `Diff` method to `ResourceModel ` that takes an `oxide.SamlIdentityProvider` and compare each attribute to make sure they're exactly the same. If some of them differ, we add a diagnostic for the attribute and return a set of `diag.Diagnostics`.

We should probably add a test case to cover this as well.

internal/provider/silo_saml_identity_provider/resource.go
```diff
@@ -392,6 +444,63 @@ func (r *Resource) Delete(
...
+		if c.configVal == c.remoteVal {
+			continue
+		}
+		diagnostics.AddAttributeError(
+			c.path,
+			"Error adopting SAML identity provider",
+			fmt.Sprintf(
+				"%q configuration does not match the state (remote=%q configuration=%q)",
+				c.path, c.remoteVal, c.configVal,
+			),
```
lgfa29, line 503:
```suggestion
			fmt.Sprintf(
				"%q value does not match remote value %q.\nUpdate your configuration file to ensure this resource matches your SAML IdP configuration.",
				c.path, c.remoteVal,
			),
```

Small suggestion to make this error message a little more precise and provide an actionable fix to users.

Image

lgfa29 (approved):
Great work getting all the weird edge cases covered.

PR #776, opened 2026-06-20
feat(oxide_instance): multicast_groups attribute

sudomateo:
The `multicast_groups` attribute was added to configure an instance's multicast groups

Closes SSE-338.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ endif
 # Acceptance test variables
 TEST_ACC_COUNT ?= 1
 TEST_ACC ?= github.com/oxidecomputer/terraform-provider-oxide/internal/...
-TEST_ACC_NAME ?= TestAcc
+TEST_ACC_NAME ?= TestAccCloudResourceInstance
 TEST_ACC_PARALLEL = 6
 TEST_ACC_OMICRON_BRANCH ?=
 TEST_ACC_SIM_HOST ?= http://localhost:12220
--- a/acctest/docker-compose.yaml
+++ b/acctest/docker-compose.yaml
@@ -6,10 +6,18 @@ services:
     environment:
       RUST_BACKTRACE: "1"
     command:
-      - "omicron-dev"
-      - "run-all"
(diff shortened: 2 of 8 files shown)


acctest/docker-compose.yaml
```diff
@@ -6,10 +6,18 @@ services:
...
+      - |
+        config=/omicron/nexus/examples/config.toml
+        if grep -q '^\[multicast\]' "$$config"; then
+          sed -i '/^\[multicast\]/,/^\[/{s/^enabled = .*/enabled = true/}' "$$config"
+        else
+          printf '\n[multicast]\nenabled = true\n' >> "$$config"
+        fi
+        exec omicron-dev run-all \
+          --nexus-config=/omicron/nexus/examples/config.toml \
+          --gateway-config=/omicron/gateway-test-utils/configs/sp_sim_config.test.toml
```
sudomateo, line 20: This should move to the `Dockerfile` instead. Also, we should update the upstream `nexus/examples/config.toml` to enable multicast by default.

Makefile
```diff
@@ -19,7 +19,7 @@ endif
 # Acceptance test variables
 TEST_ACC_COUNT ?= 1
 TEST_ACC ?= github.com/oxidecomputer/terraform-provider-oxide/internal/...
-TEST_ACC_NAME ?= TestAcc
+TEST_ACC_NAME ?= TestAccCloudResourceInstance
```
sudomateo, line 22: Didn't mean to push this. I'll correct.

```suggestion
TEST_ACC_NAME ?= TestAcc
```

sudomateo:
I'm going to chat with the team about this next week. I see the instance create and view APIs were updated to support multicast groups but the join and list APIs are listed as experimental. Additionally, Oxide has to have multicast enabled in the RSS configuration for this to even be used.

Given this split I'm reluctant to add this feature to Terraform at this time but I want to better understand what we should be doing with the `multicast_groups` attribute in the request bodies meanwhile.

sudomateo:
Closing this until multicast is actually supported in Oxide.

PR #777, opened 2026-06-22
ci: ignore errors on log collection

lgfa29:
Ignore errors on log collection to ensure we capture as many log files as possible.

Avoid problems such as what happened in https://github.com/oxidecomputer/terraform-provider-oxide/actions/runs/27942817765/job/82680160718, where we can't really tell what went wrong.

### DIFF
--- a/.github/workflows/acceptance-sim.yml
+++ b/.github/workflows/acceptance-sim.yml
@@ -114,12 +114,13 @@ jobs:
         if: always()
         continue-on-error: true
         working-directory: acctest
+        # Ignore command errors to collect as many logs as possible.
         run: |
-          docker compose exec -T omicron-dev sh -c 'tar -cf - /tmp/omicron-dev*.log' | tar -xf -
-          docker compose exec -T mitmproxy sh -c 'tar -cf - /tmp/mitmproxy.log' | tar -xf -
-          docker compose logs omicron-dev > ./tmp/omicron-dev.log
-          find .. -name terraform.log -exec cat {} \; > ./tmp/terraform.log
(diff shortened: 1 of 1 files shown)


bwagner5 (approved):
lgtm

lgfa29:
I was able to get the exact error that is happening from https://github.com/oxidecomputer/terraform-provider-oxide/actions/runs/27971513485:

```
omicron-dev-1  | thread 'main' (1) panicked at nexus/test-utils/src/starter.rs:476:62:
omicron-dev-1  | called `Result::unwrap()` on an `Err` value: failed to spawn `ddmd` (with args: ["--admin-addr", "::1", "--admin-port", "0", "--api-only", "--data-dir", "/tmp/.tmp9016lU"])
omicron-dev-1  | 
omicron-dev-1  | Caused by:
omicron-dev-1  |     No such file or directory (os error 2)
```

I think https://github.com/oxidecomputer/omicron/commit/44dc467aad3ad088288c585e3661b1b9d07d3c2b introduced new binaries that we need to copy in 
https://github.com/oxidecomputer/terraform-provider-oxide/blob/4b64ec0288d0923d8d542837ff6f0a24dc2fe324/acctest/Dockerfile#L25-L34

But I'm not sure which paths need to be included and buildomat is currently sad, so I can't run test builds right now.

lgfa29:
https://github.com/oxidecomputer/terraform-provider-oxide/actions/runs/27971513485 at least shows that this PR works in collecting logs even if one of the log collection command fails, so merging this and I will open a new one to fix CI.

PR #780, opened 2026-06-22
ci: fix simulate Omicron image

### LINKED ISSUE #778: Fix `build & test` workflow
The [build & test](https://github.com/oxidecomputer/terraform-provider-oxide/blob/main/.github/workflows/build-test.yml) workflow is failing when dispatched via its cron schedule, presumably because there's something on Omicron `main` that we're failing to account for in our container configuration. Let's get that fixed so the scheduled tests turn green.

lgfa29:
[omicron#44dc467aad3a](https://github.com/oxidecomputer/omicron/commit/44dc467aad3ad088288c585e3661b1b9d07d3c2b) introduced new dependencies to run the Omicron dev simulator: `ddmd` and `ss`, which is provided by the `iproute2` package in Debian.

Copy new binaries from the build image and install `iproute2` in the runner image.

Successful test run with these changes: https://github.com/oxidecomputer/terraform-provider-oxide/actions/runs/27979409615

Closes #778

### DIFF
--- a/acctest/Dockerfile
+++ b/acctest/Dockerfile
@@ -19,7 +19,7 @@ FROM debian:bookworm-slim
 COPY --from=builder /omicron/tools/install_runner_prerequisites.sh /tmp/
 RUN apt-get update && \
     /tmp/install_runner_prerequisites.sh -y -s && \
-    apt-get install -y curl && \
+    apt-get install -y curl iproute2 && \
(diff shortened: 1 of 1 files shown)


sudomateo (approved):

jmcarp (approved):
🚀

PR #782, opened 2026-06-25
credentials: add new function to read credentials

### LINKED ISSUE #781: Provider Function to Decode Credentials File
Let's create some `provider::oxide::credentials` function that reads an Oxide `credentials.toml` file and decodes the credentials into some data structure keyed by profile. This will allow us to use the credentials in a Terraform configuration to pass to resource that need it (e.g., the Rancher node driver).

Here's what the `credentials.toml` looks like.

```toml
[profile.dogfood]
host = "https://dogfood.sys.example.com"
token = "oxide-token-XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
token_id = "c2e35ae7-ca9b-460d-a23f-c177b9ed4388"
user = "c5bcc999-733f-44fc-b900-5a02b24446fb"

[profile.colo]
host = "https://colo.sys.example.com"
token = "oxide-token-XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
token_id = "1dfd30fd-e5c5-4b3c-809e-1fe478b9155c"
user = "febdc6fe-3ba8-4d14-b17c-5b5bdc640e92"
```

Here's how we can decode it in Terraform.

```
locals {
  oxide_credentials = provider::oxide::credentials("credentials.toml")
  
  dogfood_credentials = oxide_credentials["dogfood"]
  colo_credentials = oxide_credentials["colo"]
}
```

lgfa29:
Introduce a new function to read Oxide API credentials from a local credentials file, which can be useful when reading credentials based on profile names.

Close #781

-----

### Pull request checklist


.changelog/0.20.1.toml
sudomateo, line 1: We'll need some `.changelog/0.21.0.toml` file for the next release. I'm assuming you'll want to merge this and backport it to the v0.20 branch and then add a new changelog separately?

lgfa29:
Rebased to pull #783

internal/provider/credentials/function.go
```diff
@@ -0,0 +1,101 @@
...
+
+var _ function.Function = &Function{}
+
+const DefaultPath = ".config/oxide/credentials.toml"
+
+type credential struct {
+	Host    string `toml:"host"     tfsdk:"host"`
+	Token   string `toml:"token"    tfsdk:"token"`
+	TokenID string `toml:"token_id" tfsdk:"token_id"`
+	User    string `toml:"user"     tfsdk:"user"`
```
sudomateo, line 23: Do we even need the `tfsdk` struct tag since we're returning a `types.ObjectType`?

examples/functions/credentials/function.tf
```diff
@@ -0,0 +1,6 @@
+locals {
+  credentials = provider::oxide::credentials(null)
```
sudomateo: I thought the allow null would make this optional. I don't love passing `null` or `""` to use the default file but it's better than having two functions.

examples/functions/credentials/function.tf
sudomateo, line 1: Let's add an example showing the file name being passed.

sudomateo (commented):
Left some minor suggestions to add another example and discuss what this looks like for v0.21.0. Otherwise this looks great!

.changelog/0.20.1.toml
lgfa29: Yeah, backports are another thing that is kind of awkward with changelog. This PR will be merge to `main` and backported to `rel/v20`, and we need to make sure the changelog entry is present in both branches.

examples/functions/credentials/function.tf
lgfa29: Unfortunately not, all function args are required, so allowing `null` or an empty string is the only option.

internal/provider/credentials/function.go
lgfa29: We do because we need to transform a `credential` struct into an `ObjectType`. The `tfsdk` label is used to match struct fields to the `ObjectType` attributes.

examples/functions/credentials/function.tf
lgfa29: Added in https://github.com/oxidecomputer/terraform-provider-oxide/pull/782/changes/e2f4cbf35ba8ad8b7e85ed0122fa5f5e81398fa8. Also a good opportunity to show that `pathexpand` can be used.

.changelog/0.20.1.toml
sudomateo: Agreed. It's one of the reasons I don't love the versioned TOML files, at least in their current state.

internal/provider/credentials/function.go
sudomateo: Huh. Today I learned. Thanks!

```
╷
│ Error: Error in function call
│
│   on main.tf line 13, in output "credentials":
│   13:   value = provider::oxide::credentials(null)
│     ├────────────────
│     │ while calling provider::oxide::credentials(path)
│
│ Call to function "provider::oxide::credentials" failed: Value Conversion Error: An unexpected error was encountered trying to convert from struct value.
│ This is always an error in the provider. Please report the following to the provider developer:
│
│ error retrieving field names from struct tags: ["dogfood"]: need a struct tag for "tfsdk" on Token.
```

sudomateo (approved):

PR #794, opened 2026-06-30
oxide_ip_pool_silo_link: use composite attributes for id

sudomateo:
The `id` attribute for the `oxide_ip_pool_silo_link` resource now uses the format `IP_POOL_ID/SILO_ID`, matching the `oxide_subnet_pool_silo_link` resource.

Closes SSE-348.


internal/provider/ip_pool_silo_link/resource.go
sudomateo, line 1: We don't need a state migration for this change since `Read` uses the `ip_pool_id` and `silo_id` attributes to fetch the resource rather than `id`. This means that the next call to `Read` will update the `id` attribute for us from `IP_POOL_ID` to `IP_POOL_ID/SILO_ID`.

internal/provider/ip_pool_silo_link/resource.go
```diff
@@ -220,27 +231,27 @@ func (r *Resource) Read(
...
 		map[string]any{"success": true},
 	)
 
-	siloID := state.SiloID.ValueString()
+	ipPoolID := state.IPPoolID.ValueString()
 	idx := slices.IndexFunc(
-		links.Items,
-		func(l oxide.IpPoolSiloLink) bool { return l.SiloId == siloID },
+		pools,
+		func(p oxide.SiloIpPool) bool { return p.Id == ipPoolID },
```
lgfa29: Should we also check for `p.Name`? With https://github.com/oxidecomputer/terraform-provider-oxide/pull/795 I think it would not be possible for `ipPoolID` to be a name, but maybe there are scenarios where it could be and then the resource is removed 😬 

I think it would be harmless to add an `||` check just in case to ensure we don't run into this (famous last words 😅)

lgfa29 (approved):

internal/provider/ip_pool_silo_link/resource.go
sudomateo: It shouldn't happen but you're right the `||` is harmless. Added in https://github.com/oxidecomputer/terraform-provider-oxide/compare/a22b662f42236d8c9e271df0564fc60e65d70014..4a3649c01e3c00edf5cd7f109800ad8ee90b0aab.

PR #807, opened 2026-07-31
fix(oxide_system_ip_pools): use system_ip_pool_list api

### LINKED ISSUE #806: The `oxide_system_ip_pools` datasource does not use the `system_ip_pool_list` API
The `oxide_system_ip_pools` datasource does not use the `system_ip_pool_list` API, meaning it cannot retrieve IP pools for all silos. This prevents operators from using fleet access via one silo from read IP pools from other silos to configure them further. This was not caught when oxidecomputer/omicron#9838 landed but was reported recently by a customer. Fix it!

sudomateo:
Updated the `oxide_system_ip_pools` data source to use the `system_ip_pool_list` API, allowing it to read all the IP pools for the system. This was missed when https://github.com/oxidecomputer/omicron/issues/9838 landed but now resolved.

Resolves SSE-418.


internal/provider/subnet_pool/resource_test.go
sudomateo, line 1: This resource also uses the system API so I updated the test to verify against the system API instead.

Technically this resource should be named `oxide_system_subnet_pool`. Actually, I'm just going to log an issue for that.

lgfa29 (approved):

PR #810, opened 2026-08-05
fix(oxide_silo): mark tls_certificates[].key attribute sensitive

sudomateo:
The `oxide_silo`'s `tls_certificates[].key` attribute is now marked sensitive and its cleartext value no longer logged in debug mode, resolving Oxide Security Advisory 20260804-1.

The attribute was already write-only to keep it out of Terraform state but nothing prevented it from being printed in CLI output or logs.

Resolves SSE-421.


bwagner5 (approved):
lgtm

PR #819, opened 2026-08-09
feat(oxide_silo_saml_identity_provider): support import and mark attributes write-only

sudomateo:
…ibutes write-only

The `idp_metadata_source` and `signing_keypair` attributes are now write-only and the `oxide_silo_saml_identity_provider` resource can now be imported.

Resolves SSE-314.

### DIFF
--- a/examples/resources/oxide_silo_saml_identity_provider/resource.tf
+++ b/examples/resources/oxide_silo_saml_identity_provider/resource.tf
@@ -29,11 +29,11 @@ resource "oxide_silo_saml_identity_provider" "example" {
 
   idp_metadata_source = {
     type = "base64_encoded_xml"
-    data = base64encode(file("${path.module}/idp-metadata.xml"))
+    data = filebase64("${path.module}/idp-metadata.xml")
   }
 
   signing_keypair = {
-    private_key = base64encode(file("${path.module}/saml-key.pem"))
-    public_cert = base64encode(file("${path.module}/saml-cert.pem"))
+    private_key = filebase64("${path.module}/saml-key.der")
+    public_cert = filebase64("${path.module}/saml-cert.der")
   }
 }
--- a/internal/provider/silo_saml_identity_provider/resource.go
+++ b/internal/provider/silo_saml_identity_provider/resource.go
@@ -433,9 +433,7 @@ func (r *Resource) Read(
 	state.Description = types.StringValue(idpConfig.Description)
 	state.IdpEntityId = types.StringValue(idpConfig.IdpEntityId)
 	state.Name = types.StringValue(string(idpConfig.Name))
-	if idpConfig.PublicCert == "" {
-		state.SigningKeypair = nil
-	} else {
+	if idpConfig.PublicCert != "" {
 		state.SigningKeypair = &SigningKeypairResourceModel{
 			PrivateKey: types.StringNull(),
 			PublicCert: types.StringValue(idpConfig.PublicCert),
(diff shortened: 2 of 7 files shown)


examples/resources/oxide_silo_saml_identity_provider/resource.tf
sudomateo, line 1: We lost the changes from https://github.com/oxidecomputer/terraform-provider-oxide/pull/514 when we landed https://github.com/oxidecomputer/terraform-provider-oxide/pull/498 which is why I updated this file.

internal/provider/silo_saml_identity_provider/resource.go
```diff
@@ -405,6 +433,14 @@ func (r *Resource) Read(
 	state.Description = types.StringValue(idpConfig.Description)
 	state.IdpEntityId = types.StringValue(idpConfig.IdpEntityId)
 	state.Name = types.StringValue(string(idpConfig.Name))
+	if idpConfig.PublicCert == "" {
+		state.SigningKeypair = nil
+	} else {
```
bwagner5: nit: I don't think you need this `if`.  Won't `state.SigningKeypair` be `nil` already since it's write-only? 

Not a big deal, and I know terraform is weird with types so maybe I'm wrong too. 

```suggestion
    if idpConfig.PublicCert != "" {
		state.SigningKeypair = &SigningKeypairResourceModel{
			PrivateKey: types.StringNull(),
			PublicCert: types.StringValue(idpConfig.PublicCert),
		}
	}
```

bwagner5 (approved):
lgtm; just 1 optional comment

internal/provider/silo_saml_identity_provider/resource.go
```diff
@@ -449,17 +485,67 @@ func (r *Resource) Delete(
...
+// UpgradeState implements [resource.ResourceWithUpgradeState].
+func (r *Resource) UpgradeState(
+	ctx context.Context,
+) map[int64]resource.StateUpgrader {
+	priorSchema := r.schemaV0(ctx)
+
+	return map[int64]resource.StateUpgrader{
+		0: {
+			PriorSchema:   &priorSchema,
+			StateUpgrader: r.stateUpgraderV0ToCurrent,
```
lgfa29:
```suggestion
			StateUpgrader: r.stateUpgraderV0,
```

Minor nit to keep the name consistent with other resources. The `ToCurrent` is also kind of implicit given the [docs](https://developer.hashicorp.com/terraform/plugin/framework/resources/state-upgrade#implementing-state-upgrade-support):

> Each [resource.StateUpgrader](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/resource#StateUpgrader) implementation is expected to wholly upgrade the resource state from the prior version to the current version.

internal/provider/silo_saml_identity_provider/resource.go
```diff
@@ -449,17 +485,67 @@ func (r *Resource) Delete(
...
+		},
+	}
+}
+
+// stateUpgraderV0ToCurrent updates the v0 state to current.
+func (r *Resource) stateUpgraderV0ToCurrent(
```
lgfa29: In a similar vein, other resources declare the state upgrader in the same file as the schema (`resource_v0.go` in this case).

lgfa29 (approved):
Minor bikeshedding, but LGTM.

In theory this could be considered a breaking change since we can't read these values anymore, but I think it's unlikely anyone would be relying on this.

internal/provider/silo_saml_identity_provider/resource.go
sudomateo: I was being overly defensive here. Since this resource doesn't support updates there's really no way to get into the situation where we'd need to set it to `nil`. Addressed in 99bcf04b.

internal/provider/silo_saml_identity_provider/resource.go
sudomateo: Addressed in 99bcf04b.

internal/provider/silo_saml_identity_provider/resource.go
sudomateo: Addressed in 99bcf04b.

PR #834, opened 2026-08-12
feat: oxide_system_subnet_pool resources

sudomateo:
Deprecated these resources:

* oxide_subnet_pool
* oxide_subnet_pool_member
* oxide_subnet_pool_silo_link

Replaced them with:

* oxide_system_subnet_pool
* oxide_system_subnet_pool_member
* oxide_system_subnet_pool_silo_link

Resolves SSE-451.
Resolves SSE-452.
Resolves SSE-453.


internal/provider/system_subnet_pool_member/resource.go
```diff
@@ -0,0 +1,369 @@
...
+		return
+	}
+
+	tflog.Trace(ctx, "deleted system subnet pool member", map[string]any{
+		"pool":   state.Pool.ValueString(),
+		"subnet": state.Subnet.ValueString(),
+	})
+}
+
+func isSubnetPoolMemberNotFound(err error, subnet string) bool {
```
sudomateo: I'm doing a full string match rather than the partial `does not exist` match from the deprecated resource.

templates/guides/upgrade.md.tmpl
sudomateo, line 1: I need eyes to make sure this reads well. Especially:

* How I've changed the composite import ID from `INNER/OUTER` to `OUTER/INNER` to match hierarchy better.
* How the new resources accept name or ID.

docs/guides/upgrade.md
```diff
@@ -17,6 +17,103 @@ Refer to the
...
+ }
+```
+
+Record the pool UUID from the old resource's `id` attribute. Then remove the
+old resource from state and import the pool into the replacement resource.
+
+```shell
+terraform state rm oxide_subnet_pool.example
+terraform import oxide_system_subnet_pool.example 3e2c6e84-bed8-4c94-afc3-1032082d6a90
+```
```
lgfa29: Would a [`moved`](https://developer.hashicorp.com/terraform/language/block/moved) work here? Alternatively, maybe a [`removed`](https://developer.hashicorp.com/terraform/language/block/removed) + [`import`](https://developer.hashicorp.com/terraform/language/block/import) would work to help automate this process?

docs/guides/upgrade.md
```diff
@@ -17,6 +17,103 @@ Refer to the
...
+### Resource `oxide_subnet_pool_member`
+
+#### Deprecation
+
+The `oxide_subnet_pool_member` resource is deprecated and will be removed in
+`0.25.0`.
+
+Replace `oxide_subnet_pool_member` with `oxide_system_subnet_pool_member` and
+rename `subnet_pool_id` to `pool`. The `pool` attribute accepts a pool name or
+ID.
```
lgfa29: As discussed internally, should `pool` only accept ID?

docs/guides/upgrade.md
```diff
@@ -17,6 +17,103 @@ Refer to the
...
++resource "oxide_system_subnet_pool_member" "example" {
++  pool              = oxide_system_subnet_pool.example.id
+   subnet            = "192.0.2.0/24"
+   min_prefix_length = 24
+   max_prefix_length = 28
+ }
+```
+
+Record the pool UUID from the old resource's `subnet_pool_id` attribute
+and the member UUID from its `id` attribute. Then remove the old resource
```
lgfa29: Hum...this maybe a little annoying to track. If the `moved`/`removed`/`import` block doesn't work here, maybe use the `terraform state show` command to retrieve the correct ID to use for import?

docs/resources/subnet_pool_silo_link.md
```diff
@@ -3,20 +3,23 @@
...
 ## Example Usage
 
 ```terraform
 resource "oxide_subnet_pool_silo_link" "example" {
-  silo_id    = "1fec2c21-cf22-40d8-9ebd-e5b57ebec80f"
-  ip_pool_id = "081a331d-5ee4-4a23-ac8b-328af5e15cdc"
-  is_default = true
+  silo_id        = "1fec2c21-cf22-40d8-9ebd-e5b57ebec80f"
+  subnet_pool_id = "081a331d-5ee4-4a23-ac8b-328af5e15cdc"
+  is_default     = true
```
lgfa29: It doesn't seem like this actually changed?

internal/provider/subnet_pool/resource.go
```diff
@@ -87,7 +87,12 @@ func (r *Resource) Schema(
 	resp *resource.SchemaResponse,
 ) {
 	resp.Schema = schema.Schema{
-		MarkdownDescription: "This resource manages subnet pools. Use `oxide_subnet_pool_member` to add members to the pool.",
+		DeprecationMessage: "This resource is deprecated and will be removed in version v0.25.0 of the provider. Use oxide_system_subnet_pool instead.",
+		MarkdownDescription: shared.ReplaceBackticks(`
+This resource manages system subnet pools.
+
+!> This resource is deprecated and will be removed in version v0.25.0 of the provider. Use ''oxide_system_subnet_pool'' instead.
```
lgfa29, line 95: Hum..kind of annoying that we need to have the message here as well.

internal/provider/system_subnet_pool/resource.go
```diff
@@ -0,0 +1,353 @@
...
+				Required:    true,
+				Description: "Name of the subnet pool.",
+			},
+			"description": schema.StringAttribute{
+				Required:    true,
+				Description: "Description for the subnet pool.",
+			},
+			"ip_version": schema.StringAttribute{
+				Required:    true,
+				Description: "The IP version for this pool. All subnets in the pool must match this version.",
```
lgfa29:
```suggestion
				Description: "The IP version for this pool. All subnets in the pool must match this version. Possible values: `v4`, `v6`.",
```

internal/provider/system_subnet_pool_member/resource.go
```diff
@@ -0,0 +1,374 @@
...
+	)
+}
+
+func (r *Resource) Schema(
+	ctx context.Context,
+	_ resource.SchemaRequest,
+	resp *resource.SchemaResponse,
+) {
+	resp.Schema = schema.Schema{
+		MarkdownDescription: "This resource manages a subnet pool member using the system API.",
```
lgfa29: Reading this made we wonder if we actually need to rename these resources? Is it really relevant to users that this is a system API? Do they even know what this means?  

And it seems like they all use the same endpoint underneath. This feels like a particularly painful migration, so I think I would prefer to avoid it if possible.

internal/provider/system_subnet_pool_member/resource.go
```diff
@@ -0,0 +1,374 @@
...
+	if err != nil {
+		resp.Diagnostics.AddError("Error parsing subnet CIDR", err.Error())
+		return
+	}
+
+	body := &oxide.SubnetPoolMemberAdd{Subnet: subnet}
+	if !plan.MaxPrefixLength.IsNull() && !plan.MaxPrefixLength.IsUnknown() {
+		maxPrefixLength := int(plan.MaxPrefixLength.ValueInt64())
+		body.MaxPrefixLength = &maxPrefixLength
+	}
```
lgfa29, line 211:
```suggestion
	if !plan.MaxPrefixLength.IsNull() && !plan.MaxPrefixLength.IsUnknown() {
		body.MaxPrefixLength = oxide.NewPointer(int(plan.MaxPrefixLength.ValueInt64()))
	}
```

Minor nit-pick, feel free to ignore.

internal/provider/system_subnet_pool_member/resource.go
lgfa29: Hum...probably fine, but famous last words 😄

internal/provider/provider.go
```diff
@@ -240,8 +243,11 @@ func (p *oxideProvider) Resources(_ context.Context) []func() resource.Resource
 		sshkey.NewResource,
 		subnetpoolmember.NewResource,
 		subnetpool.NewResource,
+		systemsubnetpool.NewResource,
 		subnetpoolsilolink.NewResource,
 		switchportsettings.NewResource,
+		systemsubnetpoolmember.NewResource,
+		systemsubnetpoolsilolink.NewResource,
 		vpcfirewallrules.NewResource,
```
lgfa29: Minor nit-picking, but I would suggest keeping these sorted.

templates/guides/upgrade.md.tmpl
lgfa29: Ops, commented on the rendered file, but I can move them here if it makes more sense.

internal/provider/system_subnet_pool_member/resource.go
sudomateo: My motivation for renaming these was to make it clear that it's behind the system API and thus needs a fleet admin token to use. Really though, just IP pools and subnet pools make use of developer versus system APIs. For example, the `oxide_subnet_pool(s)` and `oxide_system_subnet_pool(s)` data sources list subnet pools in a silo or across all silos respectively. Since there will be different data sources depending on the developer/system scope, I figured moving the resource to the system scope made sense.

However, by this logic, the `oxide_silo` resource should be renamed to `oxide_system_silo`.

I don't want to be overly pedantic though. Let's chat about it more tomorrow?

sudomateo: I have the following data sources already unreleased on `main`.

```
[[features]]
title = "New data source"
description = "`oxide_system_ip_pool`"

[[features]]
title = "New data source"
description = "`oxide_system_subnet_pools`"
```

We should come to a conclusion before we release because it may affect the data sources.

internal/provider/system_subnet_pool_member/resource.go
sudomateo: When you say things like this it makes me want to go back to `does not exist` matching lol.

templates/guides/upgrade.md.tmpl
sudomateo: I went back to using the `*_id` attributes and kept the original `INNER/OUTER` import format.

docs/guides/upgrade.md
sudomateo: I'll investigate this further today.

docs/guides/upgrade.md
sudomateo: Moved back to `subnet_pool_id` and it only accepts an ID that's validated to be a UUID.

docs/resources/subnet_pool_silo_link.md
sudomateo: This was just correcting a typo that exists today: https://registry.terraform.io/providers/oxidecomputer/oxide/latest/docs/resources/subnet_pool_silo_link

image

internal/provider/subnet_pool/resource.go
sudomateo: I'll check how this is exposed in the provider. I might be able to clean this up a bit.

internal/provider/system_subnet_pool/resource.go
sudomateo: Addressed in ff2bfc1a.

internal/provider/system_subnet_pool_member/resource.go
sudomateo: Addressed in ff2bfc1a.

internal/provider/provider.go
sudomateo: Addressed in ff2bfc1a.

internal/provider/system_subnet_pool_member/resource.go
lgfa29: My main concern is that it's an implicit dependency on API version. If the error message changes, the provider may misbehave. But that seems unlikely, and also somewhat true for just `does not exist`, although there are fewer characters to change.

docs/resources/subnet_pool_silo_link.md
lgfa29: Ahh I was looking at the [wrong link](https://registry.terraform.io/providers/oxidecomputer/oxide/latest/docs/resources/ip_pool_silo_link) 😅

internal/provider/system_subnet_pool_member/resource.go
lgfa29: Chatted internally, keeping `system` in the name makes sense to highlight how some resources are expected to be used by system operators instead of developers.

internal/provider/system_subnet_pool_member/resource.go
sudomateo: I moved this back to check for `does not exist` in 1d6f526f.

docs/guides/upgrade.md
sudomateo: I updated this to support `moved` in 4aab2868.

docs/guides/upgrade.md
sudomateo: Superseded by 4aab2868.

docs/guides/upgrade.md
lgfa29: Nice! I'm glad it worked. The process seems a lot less scary now 😄

lgfa29 (approved):

internal/provider/subnet_pool/resource.go
sudomateo: Oh this is needed otherwise we would need to create a new "resource" docs template to use. I don't love it but it works.

PR #839, opened 2026-08-12
feat: oxide_system_ip_pool resources

sudomateo:
Deprecated these resources:

* oxide_ip_pool
* oxide_ip_pool_silo_link

Replaced them with:

* oxide_system_ip_pool
* oxide_system_ip_pool_range
* oxide_system_ip_pool_silo_link

Resolves SSE-274.
Resolves SSE-449.
Resolves SSE-450.


sudomateo:
Opened https://github.com/oxidecomputer/omicron/pull/11132 to update the `system_ip_pool_silo_list` API to list non-discoverable silos that are linked to the IP pool. This matches the behavior for the `system_subnet_pool_silo_list` API and allows downstream clients (e.g., Terraform) to keep their CRUD logic within a single API group.

The acceptance tests will fail until this pull request is merged.

docs/guides/upgrade.md
```diff
@@ -144,6 +148,94 @@ moved {
...
+require recording range UUIDs.
+
+```terraform
+import {
+  to = oxide_system_ip_pool_range.example
+  id = "${oxide_system_ip_pool.example.id}/172.20.18.227/172.20.18.239"
+}
+```
+
+The existing `IP_POOL_ID/RANGE_ID` import format remains supported.
```
lgfa29:
````suggestion
resource using the `id` format `IP_POOL_ID/FIRST_ADDRESS/LAST_ADDRESS`.

```terraform
import {
  to = oxide_system_ip_pool_range.example
  id = "${oxide_system_ip_pool.example.id}/172.20.18.227/172.20.18.239"
}
```
````

Minor (potentially controversial 😅) opinion, so feel free to ignore it, but I think the upgrade guide should be laser-focused on direct and actionable steps (do X, Y, Z). The sentences I removed feel like they're saying "the provider will do the right thing", which is kind of expected and doesn't add much value to the upgrade step. 

The import format is also documented in the resource, and didn't actually change, so, from the user perspective, that's not an important information.

File: docs/guides/upgrade.md [hunk 2]
```diff
@@ -144,6 +148,94 @@ moved {
...
+
+#### Deprecation
+
+The `oxide_ip_pool` resource is deprecated and will be removed in `0.25.0`. Use
+the `moved` block to migrate to the `oxide_system_ip_pool` resource and imports
+to migrate to the `oxide_system_ip_pool_range` resource.
+
+Replace `oxide_ip_pool` with `oxide_system_ip_pool` in your configuration
+and move each entry from the `ranges` attribute into its own
+`oxide_system_ip_pool_range` resource.
```
lgfa29, line 161: Nice work simplifying the migration process 🙌

internal/provider/system_ip_pool/resource.go
```diff
@@ -0,0 +1,395 @@
...
+			ctx context.Context,
+			req resource.MoveStateRequest,
+			resp *resource.MoveStateResponse,
+		) {
+			if req.SourceTypeName != "oxide_ip_pool" ||
+				req.SourceSchemaVersion != 0 ||
+				!strings.HasSuffix(
+					req.SourceProviderAddress,
+					"/oxidecomputer/oxide",
+				) {
```
lgfa29, line 107: A bit of a contrived example, but I think this means that `move` would not work for forks?

lgfa29 (approved):

internal/provider/system_ip_pool/resource.go
sudomateo: It would not, though I'd argue if they are forking they can just modify the code accordingly. I was following HashiCorp's documentation here: [https://developer.hashicorp.com/terraform/plugin/framework/resources/state-move](<https://developer.hashicorp.com/terraform/plugin/framework/resources/state-move>)

docs/guides/upgrade.md
sudomateo: Addressed in [fa4d567](https://github.com/oxidecomputer/terraform-provider-oxide/pull/839/commits/fa4d56760e93002420b6058778f443d539944a56).

Issue #140, opened 2023-05-15
Want support for the provisioner feature


askfongjojo:
The provisioner feature in Terraform allows file upload to a newly created instance which is very useful for files such as key files or other large files that are too big to be included as user_data.

The feature will require also the support of `$(self.external_ip)` to return the IP address of the newly created instance. Here is an example of how it's used in the aws provider:
```
resource "aws_instance" "jumpbox" {

  ami                         = "ami-0fcf52bcf5db7b003"
  instance_type               = "m1.medium"
  key_name                    = "${var.key_name}"
  subnet_id                   = "${var.subnet_id}"
  associate_public_ip_address = true
  user_data                   = "${data.template_file.userdata.rendered}"
  vpc_security_group_ids      = ["${aws_security_group.jumpbox_sg.id}"]
  root_block_device {
    volume_type = "standard"
  }

  tags = {
    Name = "Jumpbox"
  }

    provisioner "file" {
    source      = "~/.ssh/aws.pem"
    destination = "/home/ubuntu/.ssh/id_rsa"

    connection {
      type         = "ssh"
      user         = "ubuntu"
      host         = "${self.public_ip}"
      agent        = false
      private_key  = "${file("~/.ssh/aws.pem")}"
    }
  }

}
```

Issue #516, opened 2025-09-16
oxide_instance: attribute to read external IP addresses


sudomateo:
Currently the `oxide_instance` resource has an `external_ips` attribute that users use to specify whether they want ephemeral or floating external IPs for the instance. However, the upstream [instance_create](https://docs.oxide.computer/api/instance_create) API does not return those IP addresses back to the caller, forcing us to make a separate API call to [instance_external_ip_list](https://docs.oxide.computer/api/instance_external_ip_list). This asymmetry in the schema means that we can't just overwrite the `external_ips` attribute for the user since that'll conflict with what they specified in the configuration.

This issue is to consider updating the `oxide_instance` resource to return all the external IPs for the instance so that users don't have to use the `oxide_instance_external_ips` data source and potentially turn that data source into an apply-time data source (e.g., instance just created so data source is deferred to apply).

PR #841, opened 2026-08-20
Add `attached_external_ips` to the instance resource.

jmcarp:
In order to look up the external ips attached to an instance, we currently use the `oxide_instance_external_ips` data source, then loop over its `external_ips` to find addresses of interest:

```
one([for eip in data.oxide_instance_external_ips.instance.external_ips : eip.ip if eip.kind == "ephemeral"])
```

This is cumbersome, both because we have to use a data source for a resource that's already managed in terraform, and because the user has to know to filter addresses by kind.

This patch adds an `attached_external_ips` computed attribute to the instance resource, which provides ip addresses for all external ips, grouped by kind. Using this computed attribute, users don't need to add a data source, and don't need to remember to filter:

```
one(oxide_instance.instance.attached_external_ips.ephemeral)
```

-----

### Pull request checklist

### DIFF
--- a/.changelog/0.22.0.toml
+++ b/.changelog/0.22.0.toml
@@ -15,6 +15,10 @@ title = "New data source"
 description = "`oxide_current_user`"
 
 [[enhancements]]
+title = "`oxide_instance`"
+description = "The oxide_instance resource now exposes a read-only `attached_external_ips` attribute, which lists attached floating, ephemeral, and SNAT external IP addresses. For Terraform-managed instances, this attribute can replace the `oxide_instance_external_ips` data source. [#841](https://github.com/oxidecomputer/terraform-provider-oxide/pull/841)"
+
+[[enhancements]]
 title = "`oxide_silo_saml_identity_provider`"
 description = "The `idp_metadata_source` and `signing_keypair.private_key` attributes are now write-only. [#819](https://github.com/oxidecomputer/terraform-provider-oxide/pull/819)"
 
--- a/.changelog/0.22.0.toml
+++ b/.changelog/0.22.0.toml
@@ -15,6 +15,10 @@ title = "New data source"
 description = "`oxide_current_user`"
 
 [[enhancements]]
+title = "`oxide_instance`"
+description = "The `oxide_instance` resource now includes a read-only `attached_external_ips` that lists attached IP addresses by kind (floating, ephemeral, SNAT). This can replace the use of the `oxide_instance_external_ips` data source for terraform-managed instances. [#841](https://github.com/oxidecomputer/terraform-provider-oxide/pull/841)"
+
+[[enhancements]]
 title = "`oxide_silo_saml_identity_provider`"
 description = "The `idp_metadata_source` and `signing_keypair.private_key` attributes are now write-only. [#819](https://github.com/oxidecomputer/terraform-provider-oxide/pull/819)"
 
--- a/internal/provider/instance/resource.go
+++ b/internal/provider/instance/resource.go
@@ -64,6 +64,7 @@ type ResourceModel struct {
 	DiskAttachments           types.Set                `tfsdk:"disk_attachments"`
 	EnableJumboFrames         types.Bool               `tfsdk:"enable_jumbo_frames"`
 	ExternalIPs               *ExternalIPResourceModel `tfsdk:"external_ips"`
+	AttachedExternalIPs       types.Object             `tfsdk:"attached_external_ips"`
 	Hostname                  types.String             `tfsdk:"hostname"`
 	ID                        types.String             `tfsdk:"id"`
 	Memory                    types.Int64              `tfsdk:"memory"`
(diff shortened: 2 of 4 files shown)


jmcarp:
I chatted with @lgfa29 about this. I had vaguely remembered problems trying to implement features like this before, and he [reminded me](https://github.com/oxidecomputer/terraform-provider-oxide/pull/597#discussion_r2775150335) of the kinds of things that have gone wrong: errors correlating map/set members between plan and state, inconsistent state after apply, and maybe others. I don't have a great mental model of when these errors happen, but I believe the themes are the use of nested types like sets and maps (and thus the correlation errors) and fields that combine user-defined state and computed state. We've used data sources as an escape hatch, such as `oxide_instance_external_ips`, and exposed asymmetrical read/write fields as separate attributes on a resource, like `network_interfaces`/`attached_network_interfaces`.

I want to get a review on this, because I'm still a little confused about exactly what triggers some of these issues, but I'm also fairly confident that this branch is safe. There aren't any nested resources, so we can't have key correlation errors. And we separate the read and write config into parallel `external_ips` and `attached_external_ips` attributes, with the latter being read-only, so there's no user-intended state that can conflict with the observed state from nexus. I also ran some tests, provisioning an instance with external ips, detaching an ip out of band, and re-applying, without anything going wrong.

Relatedly, while I was researching the safety of this PR, claude pointed me at some subtle pre-existing provider bugs that I'm going to pick up as well, and regression-test. In general, I'd like to get better coverage of these edge cases, mostly so that I don't forget how they work.

internal/provider/instance/resource.go
```diff
@@ -164,6 +165,12 @@ type FloatingIPResourceModel struct {
 	ID types.String `tfsdk:"id"`
 }
 
+type AttachedExternalIPsModel struct {
+	Ephemeral []types.String `tfsdk:"ephemeral"`
+	Floating  []types.String `tfsdk:"floating"`
+	SNAT      []types.String `tfsdk:"snat"`
```
lgfa29: I wondered if we should make these proper types instead of just strings since they hold additional data, but I think that's fine for now since the IP is definitely the most relevant information (pool ID could be useful if you have `auto`, but doesn't sound _that_ useful) and the migration path doesn't seem too bad.

lgfa29 (approved):
I don't recall exactly what problems Luiz from the past found, but I ran some of the manual tests I remembered running while working on that PR and things look good here.

One question we should discus is whether to deprecate the `oxide_instance_external_ips` data source. It doesn't hurt to keep it, but we may want to nudge people towards this attribute. We can mark it as deprecated now but only remove it after several releases, since it's likely a heavily used data source.

We can probably close https://github.com/oxidecomputer/terraform-provider-oxide/issues/140 and https://github.com/oxidecomputer/terraform-provider-oxide/issues/516 with this, and we'll also need a changelog entry.

.changelog/0.22.0.toml
```diff
@@ -14,6 +14,10 @@ description = "`oxide_system_subnet_pools`"
 title = "New data source"
 description = "`oxide_current_user`"
 
+[[enhancements]]
+title = "`oxide_instance`"
+description = "The `oxide_instance` resource now includes a read-only `attached_external_ips` that lists attached IP addresses by kind (floating, ephemeral, SNAT). This can replace the use of the `oxide_instance_external_ips` data source for terraform-managed instances. [#841](https://github.com/oxidecomputer/terraform-provider-oxide/pull/841)"
```
sudomateo:
```suggestion
description = "The oxide_instance resource now exposes a read-only `attached_external_ips` attribute, which lists attached floating, ephemeral, and SNAT external IP addresses. For Terraform-managed instances, this attribute can replace the `oxide_instance_external_ips` data source. [#841](https://github.com/oxidecomputer/terraform-provider-oxide/pull/841)"
```

sudomateo (approved):
The tiniest of reviews. Very happy to have this!

.changelog/0.22.0.toml
jmcarp: Thanks, done.

internal/provider/instance/resource.go
jmcarp: Done.

jmcarp:
> One question we should discus is whether to deprecate the `oxide_instance_external_ips` data source. It doesn't hurt to keep it, but we may want to nudge people towards this attribute. We can mark it as deprecated now but only remove it after several releases, since it's likely a heavily used data source.

I was all set to do the deprecation, and was going to note that users could use the oxide_instance datasource to look up attached ips instead—but that data source doesn't actually exist. Even though it's not totally clear to me why somebody would want to look up external ips for an instance that isn't managed in terraform, it would still be a feature regression to drop the data source unless we have another path to accomplish that. I think. I'll leave oxide_instance_external_ips un-deprecated for now, and deprecate it in the future if we wind up adding an instance datasource.

sudomateo:
I think it's a good call to leave the `oxide_instance_external_ips` data source for now, but I agree that perhaps this should just be an `oxide_instance` data source instead. We should talk about how to handle the instance and disk resources and data sources overall.

PR #842, opened 2026-08-21
fix: record empty set of external ips.

jmcarp:
Currently, if a user creates an instance with `external_ips` in terraform, and then detaches them outside terraform, `terraform plan` won't detect a diff. This is incorrect: the provider should restore the detached ip addresses on the next `plan` or `apply`. This patch adds a regression test that covers ip detachment out of band, and ensures the next plan has a diff. Then we fix the failing regression test by unconditionally setting the state to the observed ips on `Read()`.

Note: we write `nil` to state rather than an empty struct, so that terraform doesn't see a spurious diff between a user-configured omitted `external_ips` and an empty struct.

-----

### Pull request checklist

### DIFF
--- a/.changelog/0.22.0.toml
+++ b/.changelog/0.22.0.toml
@@ -23,5 +23,5 @@ title = "`oxide_silo_saml_identity_provider`"
 description = "The `oxide_silo_saml_identity_provider` resource can now be imported. [#819](https://github.com/oxidecomputer/terraform-provider-oxide/pull/819)"
 
 [[bugs]]
-title = ""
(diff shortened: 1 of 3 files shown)


sudomateo (approved):
Thanks for catching this!

lgfa29 (approved):
We have lots of similar logic for other attributes, it may be worth double-checking them for the same problem. I suspect network interfaces will fail, but others may be OK because they use Terraform types.

(PR also needs a `CHANGELOG` entry)

PR #845, opened 2026-08-25
feat(oxide_vpc_internet_gateway_ip_address_attachment): new resource

sudomateo:
Added the `oxide_vpc_internet_gateway_ip_address_attachment` resource to attach IP addresses to VPC internet gateways.

Resolves SSE-65.


internal/provider/vpc_internet_gateway_ip_address_attachment/resource.go
```diff
@@ -0,0 +1,326 @@
...
+			"Error attaching IP address to internet gateway",
+			"API error: "+err.Error(),
+		)
+		return
+	}
+
+	plan.ID = types.StringValue(attachment.Id)
+	plan.Address = iptypes.NewIPAddressValue(attachment.Address)
+	plan.Name = types.StringValue(string(attachment.Name))
+	plan.Description = types.StringValue(attachment.Description)
```
lgfa29: Usually we assume that what was passed in the API request is what was recorded and returned by the API, so we only set computed or other unknown values. I think we can do the same here, or is this API different?

lgfa29 (approved):

internal/provider/vpc_internet_gateway_ip_address_attachment/resource.go
sudomateo: Addressed in 2c5582d. I always get caught up on whether `Create` should set these explicitly or just in `Read`.

lgfa29: Not a big deal, especially since I think TF does a `Read` after a `Create` anyway, but I'm mostly trying to keep things consistent.

PR #849, opened 2026-08-26
misc: use oxide.go error type for 404

sudomateo:
Removed `shared.Is404` and replaced it with `errors.Is(err, oxide.ErrHTTP404)`.


jmcarp (approved):
Nice, LGTM.

